PR #2, opened 2018-08-23
Add Swagger document for the Message service

pvallone:
[X ] This contribution adheres to [CONTRIBUTING.md](https://github.com/ni/systemlink-OpenAPI-documents/blob/master/CONTRIBUTING.md).

This pull request adds a Swagger document for the Message service. We'll treat this as the master copy of the Message service's Swagger document, for use in both SystemLink Server and SystemLink cloud. This document is a cleaned up version of the document generated from SystemLink Cloud's implementation of the Message HTTP API.

This pull requests completes a user story targeted to 18.2.

I've done basic testing of all of the routes against SystemLink Server using this document and Swagger UI.


message/message.yml
```diff
@@ -0,0 +1,412 @@
...
+        - Sessions
+      summary: Create a session
+      description: Creates a unique messaging session that can send and receive messages.
+      operationId: CreateSession
+      responses:
+        200:
+          schema:
+            $ref: '#/definitions/SessionToken'
+          description: Success
+        201:
```
spanglerco: Was this meant to be 401?

message/message.yml
```diff
@@ -0,0 +1,412 @@
...
+      description: Subscribes to a topic to receive all messages broadcast with that topic name.
+      operationId: SubscribeToTopic
+      responses:
+        200:
+          description: Success
+        401:
+          $ref: '#/responses/Unauthorized'
+        default:
+          $ref: '#/responses/Error'
+      parameters:
```
spanglerco: From a style standpoint, I would think it would be more logical to put the request information before the response information, which here would mean moving parameters above responses. Any reason in particular to put the responses above?

message/message.yml
```diff
@@ -0,0 +1,412 @@
...
+        - in: query
+          name: timeoutMilliseconds
+          description: The time to keep the connection open while waiting for a message
+          type: integer
+          required: false
```
spanglerco: Do we know if the default timeout is consistent between server and cloud? If it is, we could document it if we wanted to. We may also want to mention in the description that there's a maximum (again, we can check if there's a consistent max or not).

message/message.yml
```diff
@@ -0,0 +1,412 @@
...
+          $ref: '#/responses/Error'
+  /v1/sessions/{token}/configuration/queueSizeBytes:
+    parameters:
+      - $ref: '#/parameters/Token'
+    get:
+      tags:
+        - Message Queue
+      summary: Get the current queue size
+      description: Gets the current queue size (in bytes). This is helpful for determining the amount of memory that unread messages are consuming.
+      operationId: GetCurrentQueueSIze
```
spanglerco: SIze -> Size

message/message.yml
```diff
@@ -0,0 +1,412 @@
...
+      responses:
+        101:
+          description: Switching Protocols
+          headers:
+            Sec-WebSocket-Accept:
+              description: Used in the WebSocket opening handshake. Set to a value derived from the Sec-WebSocket-Key header.
+              type: string
+        426:
+          description: Upgrade Required
+        default:
```
spanglerco: A websocket route will never return a JSON error response. If the web service rejects the connection for any reason, the client will receive a 403 (though I don't think we have any specific failure reasons, so probably don't need to document it explicitly?).

message/message.yml
```diff
@@ -0,0 +1,412 @@
...
+        - in: header
+          name: Sec-WebSocket-Version
+          description: A value used in the WebSocket opening handshake. Indicates the protocol version of the connection.
+          type: string
+        - in: header
+          name: Sec-WebSocket-Key
+          description: A value used in the WebSocket opening handshake. Provides part of the information used by the server
+             to prove that it received a valid WebSocket opening handshake.
+          type: string
+          required: true
```
spanglerco: All of these headers are required for a successful websocket upgrade. I would say remove the default or replace them with x-example.

If we want to be pedantic about it (and I'm not sure we need to), the 101 response must also contain the Upgrade and Connection headers.

message/message.yml
```diff
@@ -0,0 +1,412 @@
...
+    required:
+      - topic
+    properties:
+      topic:
+        description: The topic name
+        type: string
+    description: The topic name
+  SessionToken:
+    type: object
+    title: SessionToken
```
spanglerco: For consistency, there should be a space in SessionToken.

spanglerco (changes requested):

message/message.yml
pvallone: Yep!

message/message.yml
pvallone: That's the order they were in in Cloud's document, but I agree it makes much more sense for the request information to be first.

message/message.yml
pvallone: From testing, they both seem to have a max timeout of ten seconds. Server seems to have a default timeout of one second, but I'm not sure what Cloud's is. I'll document these values once we know for sure.

@tschmittni: do you know what SystemLink Cloud uses as the default value for timeoutMilliseconds?

message/message.yml
pvallone: Replaced default with x-example.

I'll leave the 101 headers as is, since we're not sure if it's worth defining them.

spanglerco (dismissed):

message/message.yml
```diff
@@ -0,0 +1,415 @@
...
+    type: string
+    required: true
+
+paths:
+  /v1/sessions:
+    post:
+      tags:
+        - Sessions
+      summary: Create a session
+      description: Creates a unique messaging session that can send and receive messages.
```
jjaco2: Creates a message session with the server to send and receive messages.

message/message.yml
```diff
@@ -0,0 +1,415 @@
...
+        default:
+          $ref: '#/responses/Error'
+  /v1/sessions/{token}:
+    parameters:
+      - $ref: '#/parameters/Token'
+    delete:
+      tags:
+        - Sessions
+      summary: Delete a session
+      description: Deletes a messaging session. Messaging sessions that are not explicitly deleted are automatically cleaned up after five minutes of inactivity.
```
jjaco2: > Deletes a messaging session

Terminates the message session.

> Messaging sessions that are not explicitly deleted are automatically cleaned up after five minutes of inactivity.

If you do not delete the message session, the [message service?] deletes the session after five minutes of inactivity.

I'm curious how much we care about matching the G API. In NXG, creating a message session corresponds the Open node, and deleting a message session corresponds to the Close node. Is there a difference between creating/deleting and opening/closing?

message/message.yml
```diff
@@ -0,0 +1,415 @@
...
+security:
+  - apiKeyAuth: []
+  - basicAuth: []
+
+responses:
+  Unauthorized:
+    description: Not authorized
+    headers:
+      WWW_Authenticate:
+        description: Information for how to authenticate (only sent by SystemLink Server).
```
jjaco2: > (only sent by SystemLink Server)

(sent only from SystemLink Server)

message/message.yml
```diff
@@ -0,0 +1,415 @@
...
+      type: object
+      properties:
+        error:
+          $ref: '#/definitions/Error'
+
+parameters:
+  Token:
+    in: path
+    name: token
+    description: The unique session ID
```
jjaco2: Drop "The"

message/message.yml
```diff
@@ -0,0 +1,415 @@
...
+        default:
+          $ref: '#/responses/Error'
+  /v1/sessions/{token}/subscriptions/subscribe:
+    parameters:
+      - $ref: '#/parameters/Token'
+    post:
+      tags:
+        - Topics
+      summary: Subscribe to a topic
+      description: Subscribes to a topic to receive all messages broadcast with that topic name.
```
jjaco2: I like your description better, but I'm curious what you think of the node description we ended up with for NXG 3.0:
 
Registers an endpoint to a specific topic, or a name channel.

message/message.yml
```diff
@@ -0,0 +1,415 @@
...
+          $ref: '#/responses/Error'
+
+  /v1/sessions/{token}/messages:
+    parameters:
+      - $ref: '#/parameters/Token'
+    post:
+      tags:
+        - Message Queue
+      summary: Publish a message
+      description: Publishes a broadcast message to a specific topic. All of the topic's subscribers will receive a copy of the message.
```
jjaco2: Broadcasts a message to a topic. Subscribers of a topic receive every message published to the topic.

message/message.yml
```diff
@@ -0,0 +1,415 @@
...
+    post:
+      tags:
+        - Topics
+      summary: Unsubscribe from a topic
+      description: Unsubscribes from a topic to stop receiving messages for that topic.
+      operationId: UnsubscribeFromTopic
+      parameters:
+        - in: body
+          name: topic
+          description: The topic to unsubscribe from
```
jjaco2: Drop "The"

message/message.yml
```diff
@@ -0,0 +1,415 @@
...
+    post:
+      tags:
+        - Message Queue
+      summary: Publish a message
+      description: Publishes a broadcast message to a specific topic. All of the topic's subscribers will receive a copy of the message.
+      operationId: PublishMessage
+      parameters:
+        - in: body
+          name: topicAndMessage
+          description: The message to publish, and the topic to publish to
```
jjaco2: How does this look in practice? Does one just folow the other, or is there a specific format?

message/message.yml
```diff
@@ -0,0 +1,415 @@
...
+          description: Success
+        401:
+          $ref: '#/responses/Unauthorized'
+        default:
+          $ref: '#/responses/Error'
+    get:
+      tags:
+        - Message Queue
+      summary: Read a message from the queue
+      description: Read a single message from the queue, if any.
```
jjaco2: Returns a message from the queue. To read a message, subscribe to a topic first.

message/message.yml
jjaco2: From SystemLink Cloud's node documentation: 

> Both the SystemLink Server and the SystemLink Cloud server timeout after 10 seconds.

message/message.yml
```diff
@@ -0,0 +1,415 @@
...
+          description: No messages were available.
+        401:
+          $ref: '#/responses/Unauthorized'
+        default:
+          $ref: '#/responses/Error'
+    delete:
+      tags:
+        - Message Queue
+      summary: Flush the message queue
+      description: Flushes the queue of received messages, discarding all messages without reading them.
```
jjaco2: Clears all messages from the session queue without reading them.

message/message.yml
```diff
@@ -0,0 +1,415 @@
...
+        default:
+          $ref: '#/responses/Error'
+  /v1/sessions/{token}/configuration/queueSizeBytes:
+    parameters:
+      - $ref: '#/parameters/Token'
+    get:
+      tags:
+        - Message Queue
+      summary: Get the current queue size
+      description: Gets the current queue size (in bytes). This is helpful for determining the amount of memory that unread messages are consuming.
```
jjaco2: > Gets 

Returns

Correct me if that's not actually an accurate verb here -- I just try not to repeat the summary too much in API documentation descriptions. Could also use "obtains."

I noticed corresponding node documentation specifies it's returning the size specifically "for message queue associated with the session." Do you think that's a worthwhile distinction or can we treat it as a given?

message/message.yml
```diff
@@ -0,0 +1,415 @@
...
+        default:
+          $ref: '#/responses/Error'
+  /v1/sessions/{token}/configuration/maxQueueSizeBytes:
+    parameters:
+      - $ref: '#/parameters/Token'
+    get:
+      tags:
+        - Message Queue
+      summary: Get the maximum queue size
+      description: Gets the maximum queue size (in bytes). This is helpful for determining how much total memory is available for storing messages.
```
jjaco2: Same note as above

message/message.yml
```diff
@@ -0,0 +1,415 @@
...
+          $ref: '#/responses/Error'
+  /v1/websocket:
+    get:
+      tags:
+        - WebSocket (SystemLink Server only)
+      summary: Open a WebSocket session
+      description: Opens a persistent connection to the web server that allows bi-directional communication using a JSON
+                   formatted protocol. Once the connection is open, available actions include the WebSocket Publish Action,
+                   WebSocket Subscribe Action, and WebSocket Unsubscribe Action. The schema for these actions is defined in
+                   the *Models* section. Note that WebSocket functionality is currently only supported in SystemLink Server.
```
jjaco2: > bi-directional communication

two-way?

> a JSON formatted protocol

Looks like a lot of places online just say "JSON protocol"

> Once the connection is open, available actions include the WebSocket Publish Action, WebSocket Subscribe Action, and WebSocket Unsubscribe Action.

Do you think we need to call these actions out by their specific names, or could we say something like "once the connection is open, you can publish, subscribe, or unsubscribe?"

> The schema for these actions is defined in the *Models* section.

Refer to the *Models* section to access the schema for these actions.

> Note that WebSocket functionality is currently only supported in SystemLink Server.

SystemLink Cloud does not support WebSocket connections.

(unless you think this will confuse people who think they have Cloud when really they have Server)

message/message.yml
```diff
@@ -0,0 +1,415 @@
...
+          required: true
+          x-example: websocket
+        - in: header
+          name: Connection
+          type: string
+          required: true
+          x-example: Upgrade
+        - in: header
+          name: Sec-WebSocket-Version
+          description: A value used in the WebSocket opening handshake. Indicates the protocol version of the connection.
```
jjaco2: Protocol version of the connection to [the? a?] WebSocket.

message/message.yml
```diff
@@ -0,0 +1,415 @@
...
+          type: string
+          required: true
+          x-example: 13
+        - in: header
+          name: Sec-WebSocket-Key
+          description: A value used in the WebSocket opening handshake. Provides part of the information used by the server
+             to prove that it received a valid WebSocket opening handshake.
```
jjaco2: Value that confirms a valid WebSocket opening handshake.

What determines this? And if this is only part of the information, what's the rest? Is it the stuff above or something else?

message/message.yml
```diff
@@ -0,0 +1,415 @@
...
+          type: string
+          required: true
+          x-example: dGhlIHNhbXBsZSBub25jZQ==
+      responses:
+        101:
+          description: Switching Protocols
+          headers:
+            Sec-WebSocket-Accept:
+              description: Used in the WebSocket opening handshake. Set to a value derived from the Sec-WebSocket-Key header.
```
jjaco2: Value derived from the Sec-WebSocket-Key header to [accept?] a WebSocket opening handshake.

This might not be relevant, but how is it derived from that?

message/message.yml
```diff
@@ -0,0 +1,415 @@
...
+        426:
+          description: Upgrade Required
+
+definitions:
+  Error:
+    description: Contains error information
+    type: object
+    properties:
+      name:
+        description: The string error code
```
jjaco2: Here and throughout these: drop "The" and "A"

(really minor. can CAR this for later if you want)

message/message.yml
```diff
@@ -0,0 +1,415 @@
...
+        description: Numeric error code
+        type: integer
+      resourceType:
+        description: The type of resource associated with the error
+        type: string
+      resourceId:
+        description: The identifier of the resource associated with the error
+        type: string
+      message:
+        description: The filled in error message
```
jjaco2: > filled in

complete?

message/message.yml
```diff
@@ -0,0 +1,415 @@
...
+    required:
+      - maxQueueSizeBytes
+    properties:
+      maxQueueSizeBytes:
+        description: The maximum queue size (in bytes)
+        type: integer
+    description: The maximum queue size (in bytes)
+  WebSocketPublish:
+    type: object
+    title: WebSocket Publish Action
```
jjaco2: Could "Action" be left off these and still make sense?

message/message.yml
```diff
@@ -0,0 +1,415 @@
...
+    get:
+      tags:
+        - Message Queue
+      summary: Read a message from the queue
+      description: Read a single message from the queue, if any.
+      operationId: ReadMessage
+      parameters:
+        - in: query
+          name: timeoutMilliseconds
+          description: The time to keep the connection open while waiting for a message
```
jjaco2: Amount of time, in milliseconds, to keep the messaging connection open.

jjaco2 (commented):
Edits for clarity and style plus a few notes about how much (if at all) we care about  matching corresponding nodes in NXG.

message/nimessage.yml
```diff
@@ -0,0 +1,415 @@
...
+    name: token
+    description: The unique session ID
+    type: string
+    required: true
+
+paths:
+  /v1/sessions:
+    post:
+      tags:
+        - Sessions
```
BKnight760: All tags should be lowercase.

message/nimessage.yml
```diff
@@ -0,0 +1,415 @@
...
+    description: The topic name
+  SessionToken:
+    type: object
+    title: Session Token
+    required:
+      - token
+    properties:
+      token:
+        description: A unique session ID
+        type: string
```
BKnight760: This should include an "example: <GUID>" so that it renders an actual token id instead of just
{
  "token": "string"
}

This should be done throughout other objects in this review so that the rendered swagger UI (as well as client imports) generate output that looks real.

message/nimessage.yml
```diff
@@ -0,0 +1,415 @@
...
+    get:
+      tags:
+        - Message Queue
+      summary: Read a message from the queue
+      description: Read a single message from the queue, if any.
+      operationId: ReadMessage
+      parameters:
+        - in: query
+          name: timeoutMilliseconds
+          description: The time to keep the connection open while waiting for a message
```
BKnight760: May be good to note that this value may be overridden by a maximum value that the webserver administrator allows.

message/nimessage.yml
```diff
@@ -0,0 +1,415 @@
...
+        default:
+          $ref: '#/responses/Error'
+    delete:
+      tags:
+        - Message Queue
+      summary: Flush the message queue
+      description: Flushes the queue of received messages, discarding all messages without reading them.
+      operationId: DeleteMessages
+      responses:
+        200:
```
BKnight760: This probably should have been a 204.  It would require a change on the webservice itself to fix it.

message/nimessage.yml
```diff
@@ -0,0 +1,415 @@
...
+    type: object
+    title: Maximum Queue Size
+    required:
+      - maxQueueSizeBytes
+    properties:
+      maxQueueSizeBytes:
+        description: The maximum queue size (in bytes)
+        type: integer
+    description: The maximum queue size (in bytes)
+  WebSocketPublish:
```
BKnight760: This seems odd to have 3 different "Action" objects that define single-element ENUMs.  It may be better to have a single "WebSocket Action" object, and then have an "action" ENUM that has ["publish", "subscribe", "unsubscribe"], and then have the description explain how the different actions behave.  That would mirror the actual implementation.

BKnight760 (changes requested):

message/message.yml
pvallone: That's a good point about the G API. There's no difference between creating/deleting and opening/closing. My feeling is that clients of REST APIs are used to the concepts of creating and deleting resources, so it makes sense to describe these routes as creating and deleting resources, but I don't feel too strongly. I suspect that 'opening' and 'closing' stuff is a G terminology convention.. is that the case, @spanglerco?

message/message.yml
pvallone: I think the endpoint they're referring to is the node that gets dropped, so that seems reasonable. I think both are okay in their respective contexts.

message/message.yml
pvallone: One just follows the other, and the order doesn't matter. They're specified in a HTTP request body in JSON format. An example is below.
```
{
  "topic": "food",
  "message": "donuts are yummy"
}
```

message/message.yml
pvallone: I think 'Returns' is fine.

I like including the fact that we're dealing with the queue associated with the session. I prefer that to just talking about the 'current' session. How do you feel about: "Returns the size (in bytes) of the message queue associated with this session. This is helpful for determining the amount of memory that unread messages are consuming."

message/message.yml
pvallone: How do you feel about: "Returns the maximum size (in bytes) of the message queue associated with this session. This is helpful for determining how much total memory is available for storing messages."

message/message.yml
pvallone: > Do you think we need to call these actions out by their specific names, or could we say something like "once the connection is open, you can publish, subscribe, or unsubscribe?"
Yeah, I think that's fine. We call them actions in the next sentence, and the names are discoverable enough.

message/message.yml
pvallone: Hm, how about "Protocol version of the WebSocket connection". "WebSocket connection" is a phrase which is used a lot in the WebSocket RFC: https://tools.ietf.org/html/rfc6455.

message/message.yml
pvallone: Clients are supposed to send a random base64-encoded byte string. The rest of the stuff is also part of the handshake, yes - the Version, Upgrade, and Connection headers.

My understanding of this header wasn't very good when I first wrote the description. How do you feel about this:
"Randomly selected sixteen-byte base64-encoded string. Used to confirm a valid WebSocket handshake. Must be selected randomly for each new connection."

We're sort of paraphrasing the WebSocket RFC here (https://tools.ietf.org/html/rfc6455), and that's the resource folks are going to rely on when they want to figure out how to do standard WebSocket stuff, so we could always just point folks to that documentation in our description.

message/message.yml
pvallone: The string "258EAFA5-E914-47DA-95CA-C5AB0DC85B11" is appended to the value of Sec-WebSocket-Key, and then the server takes the SHA1 hash of the result. That's defined as part of the standard, so I'm not sure if it's worth it for us to duplicate that information in our documentation, but it's up to you.

message/message.yml
pvallone: Removed leading "The" and "A" from the model descriptions

message/message.yml
pvallone: I'd rather leave it on, as "action" has a very specific meaning in WebSockets. Swagger doesn't support documenting WebSocket APIs, so we have to be as descriptive as possible in our WebSocket action models, and I want to make sure folks can figure out which of our models correspond to our WebSocket actions.

message/nimessage.yml
pvallone: Here, the strings are used to group the routes in Swagger UI, and I don't think we want lowercase for that.

message/nimessage.yml
pvallone: Added a note. Server allows for a ten second maximum by default, and we're waiting on Thomas to tell us what Cloud's maximum is.

message/nimessage.yml
pvallone: I filed a CAR: http://nippm.natinst.com/itg/web/knta/crt/RequestDetail.jsp?REQUEST_ID=711012

message/nimessage.yml
pvallone: Dropping per our offline conversation

message/message.yml
spanglerco: Open/Close is essentially the default term to use in LabVIEW, but you can also find:
    - Create/Destroy for things that you are always creating and never reusing (like user events)
    - Obtain/Release for things that you may create or reuse if it already exists (like queues)
    - Open/Close for resources (like files)

In retrospect, we probably should have gone with Create/Destroy for the message VIs. Create/Delete is probably the correct REST term to use.

message/nimessage.yml
spanglerco: The example "pet store" Swagger file that appears when you first open the editor also uses lowercase tags, so we assume that's the convention.

message/nimessage.yml
spanglerco: The same would apply to subscribe, unsubscribe, and publish.

pvallone: Updated the CAR

spanglerco (dismissed):

message/message.yml
jjaco2: Perfect!

message/message.yml
jjaco2: Cool. In that case, how about:

After you open a connection, you can publish messages and subscribe or unsubscribe to topics [through the WebSocket connection?]

message/message.yml
jjaco2: Sounds good.

message/message.yml
jjaco2: I think that looks good.

message/message.yml
jjaco2: Hah, yeah that's not super helpful. Never mind!

message/message.yml
pvallone: Sure. My feeling is that we don't need to specify that this is being done through the WebSocket connection, as the summary and the tag contextualize this as a WebSocket connection.

message/nimessage.yml
pvallone: Added examples to all of the objects

jjaco2 (approved):

BKnight760 (approved):

PR #3, opened 2018-08-28
Add Swagger document for the Test Monitor service

BKnight760:

This pull request adds a Swagger document for the Test Monitor service. This will be the master copy of the Test Monitor service's Swagger document, for use in SystemLink Server. This document is used as the source for the webservice interface which is implemented by the Test Monitor service.

This pull request gives higher visibility to the Test Monitor HTTP API, and ensures that the implentation matches the fully documented options for the webservice.

Existing automated tests have been run against the swagger interface, and all tests pass.

### DIFF
--- a/test-monitor/nitestmonitor.yml
+++ b/test-monitor/nitestmonitor.yml
@@ -29,7 +29,7 @@ definitions:
     type: object
     properties:
       name:
-        description: The string error code
+        description: String error code
         type: string
         example: Skyline.OneOrMoreErrorsOccurred
       code:
@@ -37,15 +37,15 @@ definitions:
         type: integer
         example: -251040
       resourceType:
-        description: The type of resource associated with the error
+        description: Type of resource associated with the error
         type: string
         example: TestResult
       resourceId:
-        description: The identifier of the resource associated with the error
+        description: Identifier of the resource associated with the error
         type: string
         example: 4afb2ce3741fe11d88838cc9
       message:
-        description: The filled in error message
+        description: Filled in error message
         type: string
         example: >-
           One or more errors occurred. See the contained list for details of
@@ -82,7 +82,7 @@ definitions:
           message: Invalid Id.
           args: []
   Operation:
-    description: An operation provided by the API
+    description: Operation provided by the API
     type: object
     properties:
       available:
@@ -90,12 +90,13 @@ definitions:
         description: Whether the operation is available to the caller
       version:
         type: integer
-        description: The version of the available operation
+        description: Version of the available operation
     required: [available]
     example:
       available: true
       version: 1
   V1Operations:
+    title: V1 Operations
     description: V1 operations
     type: object
     properties:
@@ -145,13 +146,14 @@ definitions:
           deleteManySteps:
             $ref: '#/definitions/Operation'
   StatusObject:
-    description: A status object
+    title: Status Object
+    description: Status object comprised of a type and display name
     type: object
     required:
       - statusType
     properties:
       statusType:
-        description: The status type
+        description: Status type enum
         type: string
         enum:
           - LOOPING
@@ -167,11 +169,12 @@ definitions:
           - TIMED_OUT
         example: PASSED
       statusName:
-        description: The status name
+        description: Status name
         type: string
         example: Passed
   ResultField:
-    description: A result field enum
+    title: Result Field
+    description: Result field enum
     type: string
     enum:
       - ID
@@ -187,7 +190,8 @@ definitions:
       - PROPERTIES
       - FILE_IDS
   StepField:
-    description: A step field enum
+    title: Step Field
+    description: Step field enum
     type: string
     enum:
       - NAME
@@ -205,7 +209,8 @@ definitions:
       - DATA_MODEL
       - DATA
   ResultSortDefinitionObject:
-    description: A result sort definition object
+    title: Result Sort Definition
+    description: Defines the field and direction for sorting test results.
     type: object
     required:
       - field
@@ -216,7 +221,8 @@ definitions:
         type: boolean
         default: false
   StepSortDefinitionObject:
-    description: A step sort definition object
+    title: Step Sort Definition
+    description: Defines the field and direction for sorting test steps.
     type: object
     required:
       - field
@@ -227,14 +233,15 @@ definitions:
         type: boolean
         default: false
   TimeQueryObject:
-    description: A time query object
+    title: Time Query
+    description: Time query object
     type: object
     required:
       - operation
       - comparisonValue
     properties:
       operation: 
-        description: The operation enum value
+        description: Operation enum
         type: string
         enum:
           - EQUAL
@@ -244,26 +251,26 @@ definitions:
           - LESS_THAN_OR_EQUAL
           - GREATER_THAN_OR_EQUAL
       comparisonValue: 
-        description: An ISO-8601 formatted timestamp value to compare 
+        description: ISO-8601 formatted timestamp value to compare
         type: string
         format: iso-date-time
         example: '2017-11-14T15:41:06.106Z'
   TestResultRequestObject:
-    description: A test result request object
+    title: Test Result Request
     type: object
     properties:
       programName:
-        description: The program name
+        description: Program name
         type: string
         example: My Program Name
       status:
         $ref: '#/definitions/StatusObject'
       systemId:
-        description: The system id
+        description: System id
         type: string
         example: my-system
       properties:
-        description: The test result properties
+        description: Test result properties
         type: object
         additionalProperties:
           type: string
@@ -278,32 +285,32 @@ definitions:
           - keyword1
           - keyword2
       serialNumber:
-        description: The serial number
+        description: Serial number
         type: string
         example: 123-456
       operator:
-        description: The operator
+        description: Operator
         type: string
         example: admin
       fileIds:
-        description: The file ids attached to the test result
+        description: Array of file ids attached to the test result
         type: array
         items:
           type: string
         example: df1b63da-4fd8-4466-b721-5a28de036f56
       startedAt:
-        description: An ISO-8601 formatted timestamp indicating when the result was started
+        description: ISO-8601 formatted timestamp indicating when the test result began
         type: string
         format: iso-date-time
         example: '2018-05-07T18:58:05.219692Z'
       totalTimeInSeconds:
-        description: The total run-time of the test in seconds
+        description: Total run-time of the test in seconds
         type: number
         format: double
         default: 0
         example: 29.9
   TestResultResponseObject:
-    description: A test result response object
+    title: Test Result Response
     type: object
     required:
       - id
@@ -311,37 +318,37 @@ definitions:
       status:
         $ref: '#/definitions/StatusObject'
       startedAt:
-        description: An ISO-8601 formatted timestamp indicating when the result was started
+        description: ISO-8601 formatted timestamp indicating when the test result began
         type: string
         format: iso-date-time
         example: '2018-05-07T18:58:05.219692Z'
       updatedAt:
-        description: An ISO-8601 formatted timestamp indicating when the result was last updated
+        description: ISO-8601 formatted timestamp indicating when the result was last updated
         type: string
         format: iso-date-time
         example: '2018-05-09T15:07:42.527921Z'
       programName:
-        description: The program name
+        description: Program name
         type: string
         example: My test program
       id:
-        description: The id of the test result
+        description: Id of the test result
         type: string
         example: df1b63da-4fd8-4466-b721-5a28de036f56
       systemId:
-        description: The id of the system
+        description: Id of the system
         type: string
         example: sedwards-dt2
       operator:
-        description: The name of the operator running the test
+        description: Name of the operator running the test
         type: string
         example: admin
       serialNumber:
-        description: The sequential number of the device under test
+        description: Sequential number of the device under test
         type: string
         example: abc-123
       totalTimeInSeconds:
-        description: The total run-time of the test in seconds
+        description: Total run-time of the test in seconds
         type: number
         format: double
         example: 29.9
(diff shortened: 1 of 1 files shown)


test-monitor/nitestmonitor.yml
```diff
@@ -0,0 +1,1277 @@
...
+  cookieAuth:
+    type: apiKey
+    in: header
+    name: Cookie
+security:
+  - basicAuth: []
+  - cookieAuth: []
+x-ni-routing-key: Skyline.TestMonitor
+definitions:
+  Error:
```
pvallone, line 27: Jenna suggested some changes to my Error definition, so you should probably try to incorporate those, as these should be pretty similar. At some point in the future we could look into single-sourcing the error type, so we don't all have a copy of it in our docs. I don't think it needs to happen now, given that we're trying to release.

test-monitor/nitestmonitor.yml
```diff
@@ -0,0 +1,1277 @@
...
+          description: Not Found
+          schema:
+            $ref: '#/definitions/Error'
+  /v1/query-results:
+    post:
+      operationId: query-results
+      x-ni-operation: queryResults
+      x-ni-privilege: Read
+      summary: Queries test results
+      description: Queries test results
```
pvallone: If we think it isn't intuitive, we could document that an empty body returns everything.

test-monitor/nitestmonitor.yml
```diff
@@ -0,0 +1,1277 @@
...
+        error:
+          $ref: '#/definitions/Error'
+paths:
+  /:
+    get:
+      operationId: RootEndpoint
+      x-ni-request-all-privileges: true
+      summary: API information
+      description: Returns information about API versions and available operations.
+      tags: [versioning]
```
pvallone: Do you think we should try to be consistent with respect to how we order these properties across documents? In the message document I define tags, summary, description, operationId, and then responses, in that order. It might not be a big deal if we're at least consistent within each individual document.

test-monitor/nitestmonitor.yml
```diff
@@ -0,0 +1,1277 @@
...
+      responses:
+        200:
+          $ref: '#/responses/StepsDeletePartialSuccessResponse'
+        204:
+          description: The resource was deleted successfully.
+        401:
+          $ref: '#/responses/Unauthorized'
+        default:
+          $ref: '#/responses/Error'
+  /v1/ping:
```
pvallone, line 1243: Have we shipped documentation for this route to customers? If we haven't, I would think we can just leave this route undocumented, as it's deprecated.

pvallone (dismissed):
Light review given Paul and Alex reviewed this already

test-monitor/nitestmonitor.yml
spanglerco: We are using codegen for this service, so it has to be documented unless we remove it from the service (which we probably should anyway).

test-monitor/nitestmonitor.yml
```diff
@@ -0,0 +1,1277 @@
...
+security:
+  - basicAuth: []
+  - cookieAuth: []
+x-ni-routing-key: Skyline.TestMonitor
+definitions:
+  Error:
+    description: Contains error information
+    type: object
+    properties:
+      name:
```
jjaco2, line 31: Non-essential but worth noting for consistency in the future: drop leading articles like "The" and "A" throughout all  first sentences of descriptions.

test-monitor/nitestmonitor.yml
```diff
@@ -0,0 +1,1277 @@
...
+    example:
+      available: true
+      version: 1
+  V1Operations:
+    description: V1 operations
+    type: object
+    properties:
+      operations:
+        description: >-
+          Available operations in the v1 version of the API:
```
jjaco2, line 90: Could we remove "The ability to" in all the items below, or am I misunderstanding something?

test-monitor/nitestmonitor.yml
```diff
@@ -0,0 +1,1277 @@
...
+          deleteManySteps:
+            $ref: '#/definitions/Operation'
+  StatusObject:
+    description: A status object
+    type: object
+    required:
+      - statusType
+    properties:
+      statusType:
+        description: The status type
```
jjaco2: Do we need to specify this is an enum like we do for ResultField?

test-monitor/nitestmonitor.yml
```diff
@@ -0,0 +1,1277 @@
...
+      - PATH
+      - PATH_IDS
+      - STATUS
+      - TOTAL_TIME_IN_SECONDS
+      - STARTED_AT
+      - UPDATED_AT
+      - DATA_MODEL
+      - DATA
+  ResultSortDefinitionObject:
+    description: A result sort definition object
```
jjaco2: What does this mean? I get that this isn't the most essential part of this doc, but if we're not expecting the user to necessarily have experience with HTTP APIs, and this isn't really an industry term, it could be worth rephrasing in a way that doesn't repeat the object name.

test-monitor/nitestmonitor.yml
```diff
@@ -0,0 +1,1277 @@
...
+  TestResultResponseObject:
+    description: A test result response object
+    type: object
+    required:
+      - id
+    properties:
+      status:
+        $ref: '#/definitions/StatusObject'
+      startedAt:
+        description: An ISO-8601 formatted timestamp indicating when the result was started
```
jjaco2: > indicating when the result was started

Maybe "indicating the start of the timeframe for the result" or "indicating when the result began"

test-monitor/nitestmonitor.yml
```diff
@@ -0,0 +1,1277 @@
...
+          - keyword2
+      properties:
+        description: The test result properties
+        type: object
+        additionalProperties:
+          type: string
+        example:
+          key1: value1
+      fileIds:
+        description: A list of attached file ids
```
jjaco2: A list of file ids attached to the result

test-monitor/nitestmonitor.yml
```diff
@@ -0,0 +1,1277 @@
...
+        type: array
+        items:
+          type: string
+        example: []
+  TestResultUpdateRequestObject:
+    description: A test result update request object
```
jjaco2: Same note as for ResultSortDefinitionObject, and really any object descriptions that match the name verbatim: is there a way to add value to users who might not already be familiar with this domain by rephrasing the name?

Maybe "Object representing test result information for [however this object gets used, I'm not sure]"

Same goes for TestResultQueryObject, too

test-monitor/nitestmonitor.yml
```diff
@@ -0,0 +1,1277 @@
...
+          type: string
+          example:
+            key1: value1
+      updatedAtQuery:
+        description: A time query object
+        type: array
+        items:
+          $ref: '#/definitions/TimeQueryObject'
+      startedAtQuery:
+        description: A time query object
```
jjaco2: Since this is the same as the previous description, is there a way to differentiate the two?

test-monitor/nitestmonitor.yml
```diff
@@ -0,0 +1,1277 @@
...
+          $ref: '#/definitions/TimeQueryObject'
+      fileIds:
+        description: An array of file ids
+        type: array
+        items:
+          type: string
+        example:
+          - 5982002b5d67371430c80e73
+      statuses:
+        description: An array of resulkt statuses
```
jjaco2: > resulkt

test-monitor/nitestmonitor.yml
```diff
@@ -0,0 +1,1277 @@
...
+        type: array
+        items:
+          $ref: '#/definitions/StatusObject'
+      sortBy:
+        type: array
+        items:
+          $ref: '#/definitions/ResultSortDefinitionObject'
+  StepIdResultIdPair:
+    description: A pair of step id and result id
```
jjaco2: Step id and result id

test-monitor/nitestmonitor.yml
```diff
@@ -0,0 +1,1277 @@
...
+    required:
+      - steps
+    properties:
+      steps:
+        description: An array of test step id and result id pairs to delete
+        type: array
+        items:
+          $ref: '#/definitions/StepIdResultIdPair'
+  TestStepQueryObject:
+    description: A test step query object
```
jjaco2: Maybe "Object for querying a test step"

test-monitor/nitestmonitor.yml
```diff
@@ -0,0 +1,1277 @@
...
+        example: 1f05c3c8-7a8c-4ebb-a87f-7a60d5dd8938
+      children:
+        description: Nested child steps
+        type: array
+        items:
+          $ref: '#/definitions/TestStepRequestObject'
+      data:
+        $ref: '#/definitions/StepDataObject'
+      dataModel:
+        description: The step data model
```
jjaco2: Data model for the step

test-monitor/nitestmonitor.yml
```diff
@@ -0,0 +1,1277 @@
...
+    required:
+      - steps
+    properties:
+      steps:
+        description: An array of test steps to create
+        type: array
+        items:
+          $ref: '#/definitions/TestStepRequestObject'
+  TestStepResponseObject:
+    description: A test step response object
```
jjaco2: Same note about adding value for less HTTP API-savvy users by rephrasing and/or adding (very brief) context

test-monitor/nitestmonitor.yml
```diff
@@ -0,0 +1,1277 @@
...
+        - results
+        - totalCount
+      properties:
+        results:
+          description: An array of test results
+          type: array
+          items:
+            $ref: '#/definitions/TestResultResponseObject'
+        totalCount:
+          description: The total count of matching test results
```
jjaco2: > count of

number of

test-monitor/nitestmonitor.yml
```diff
@@ -0,0 +1,1277 @@
...
```
jjaco2: Same note as before (count -> number) Not sure if single-sourced

test-monitor/nitestmonitor.yml
```diff
@@ -0,0 +1,1277 @@
...
+          type: integer
+  StepsDeletePartialSuccessResponse:
+    description: Delete Test Steps Partial Success Response
+    schema:
+      description: Delete Test Steps Partial Success Response
+      title: StepsDeletePartialSuccessResponse
+      type: object
+      properties:
+        steps:
+          description: An array of test step id result id paris that were deleted
```
jjaco2: > paris

test-monitor/nitestmonitor.yml
```diff
@@ -0,0 +1,1277 @@
...
+          type: array
+          items:
+            $ref: '#/definitions/StepIdResultIdPair'
+        failed:
+          description: An array of test step id result id paris that failed to delete
```
jjaco2: > paris

test-monitor/nitestmonitor.yml
```diff
@@ -0,0 +1,1277 @@
...
+          $ref: '#/responses/Unauthorized'
+        default:
+          $ref: '#/responses/Error'
+  /v1/results:
+    put:
+      operationId: update-results
+      x-ni-operation: updateResults
+      x-ni-privilege: Write
+      summary: Update test results
+      description: Update test results
```
jjaco2: Updates existing test results

Alternatively, you could delete the summary from all these routes, since they're usually the same as the description, and IIRC, @pvallone only used descriptions.

test-monitor/nitestmonitor.yml
```diff
@@ -0,0 +1,1277 @@
...
+        401:
+          $ref: '#/responses/Unauthorized'
+        default:
+          $ref: '#/responses/Error'
+    post:
+      operationId: create-results
+      x-ni-operation: createResults
+      x-ni-privilege: Write
+      summary: Create test results
+      description: Create test results
```
jjaco2: Creates new test results

(or cut the summary)
(this note applies to all the summaries in this section)

test-monitor/nitestmonitor.yml
```diff
@@ -0,0 +1,1277 @@
...
+        - results
+      parameters:
+        - in: path
+          name: resultId
+          description: The id of the test result to delete
+          type: string
+          required: true
+        - in: query
+          name: deleteSteps
+          description: Delete the test steps associated with the test result
```
jjaco2: Indicates whether to delete the test steps associated with the test result

test-monitor/nitestmonitor.yml
```diff
@@ -0,0 +1,1277 @@
...
+              - ids
+            properties:
+              ids:
+                description: An array of test result ids to delete
+                type: array
+                items:
+                  type: string
+                  example: 1f05c3c8-7a8c-4ebb-a87f-7a60d5dd8938
+              deleteSteps:
+                description: Delete the test steps associated with the test result
```
jjaco2: "Indicates whether to...."

test-monitor/nitestmonitor.yml
jjaco2: Should we note it's deprecated outside of the tags section, like in the description?

jjaco2 (commented):
Mostly minor things plus some opportunities to briefly clarify some concepts/terminology for more novice users.

test-monitor/nitestmonitor.yml
BKnight760: This one happens to be odd, because every operation is explicitly identified here.  It may make sense to combine these into more general operations.  Then the explanation of specific requests will make sense.  Since that requires a backend service change, I'll defer to doing that in a separate change.

test-monitor/nitestmonitor.yml
BKnight760: I've marked it deprecated in the description, although I agree we should make any required changes to the UI and drop it from the API entirely.  I'll do that as a separate change since it requires a backend code change.

pvallone (approved):

PR #4, opened 2018-09-05
Add a Swagger document for the Tag HTTP API

pvallone:
[X] This contribution adheres to [CONTRIBUTING.md](https://github.com/ni/systemlink-OpenAPI-documents/blob/master/CONTRIBUTING.md).

This pull request adds a Swagger document for the Tag service. We'll treat this as the master copy of the Tag service's Swagger document, for use in both SystemLink Server and SystemLink Cloud. I based this document on SystemLink Cloud's Swagger document, available at api.systemlinkcloud.com, but a lot of changes have been made.

This pull requests completes a user story targeted to 18.2.

I've tested some of the routes against SystemLink Server, but I concede that I need to do a lot more testing against both SystemLink Server and Cloud. Since this document is so large, I figured I'd put the review up sooner rather than later. I'll do more testing while you folks are working through it.


tag/nitag.yml
```diff
@@ -0,0 +1,1814 @@
...
+          items:
+            type: string
+
+paths:
+  /v2/subscriptions:
+    post:
+      tags:
+        - subscriptions
+      summary: Create a subscription
+      description: Create a subscription for a list of tags to get notifications when the tags change.
```
jjaco2: Is this the same as subscribing to a tag? Or does this just make subscribing an option?

tag/nitag.yml
```diff
@@ -0,0 +1,1814 @@
...
+          description: Bad Request
+        401:
+          $ref: '#/responses/Unauthorized'
+        default:
+          $ref: '#/responses/Error'
+    get:
+      tags:
+        - subscriptions
+      summary: Get active subscription IDs
+      description: Get a list of all active subscription IDs
```
jjaco2: Returns

(both to make verb tense match industry standard and to reword the operation in a way that might add value)

tag/nitag.yml
jjaco2: Creates

tag/nitag.yml
```diff
@@ -0,0 +1,1814 @@
...
+      responses:
+        200:
+          description: Success
+          schema:
+            description: Get subscriptions response
+            type: object
+            title: Get Subscriptions Response
+            properties:
+              subscriptionIds:
+                description: List of all the active subscription IDs
```
jjaco2, line 134: I know this isn't the first time this is included, but what does it mean for a subscription to be active?

tag/nitag.yml
```diff
@@ -0,0 +1,1814 @@
...
+        401:
+          $ref: '#/responses/Unauthorized'
+        default:
+          $ref: '#/responses/Error'
+  /v2/subscriptions/{id}:
+    get:
+      tags:
+        - subscriptions
+      summary: Get a subscription
+      description: Get a subscription's metadata
```
jjaco2: Returns

tag/nitag.yml
```diff
@@ -0,0 +1,1814 @@
...
+          $ref: '#/responses/Unauthorized'
+        404:
+          description: Subscription with the given ID was not found
+        default:
+          $ref: '#/responses/Error'
+    delete:
+      tags:
+        - subscriptions
+      summary: Delete a subscription
+      description: Delete a specific subscription.
```
jjaco2: Deletes

Also, does this only work for one subscription at a time?

tag/nitag.yml
```diff
@@ -0,0 +1,1814 @@
...
+      operationId: DeleteSubscription
+      parameters:
+        - in: path
+          name: id
+          description: Subscription ID
+          type: string
+          required: true
+      responses:
+        204:
+          description: Successfully deleted
```
jjaco2: Subscription successfully deleted

tag/nitag.yml
```diff
@@ -0,0 +1,1814 @@
...
+        404:
+          description: Subscription with the given ID was not found
+        default:
+          $ref: '#/responses/Error'
+  /v2/subscriptions/{id}/heartbeat:
+    put:
+      tags:
+        - subscriptions
+      summary: Update the heartbeat timer for a subscription
+      description: Update the heartbeat timer for the specified subscription
```
jjaco2: What does this mean? If not every user will be familiar with this term (or maybe they will be, idk), maybe elaborate or rephrase in the description.

tag/nitag.yml
```diff
@@ -0,0 +1,1814 @@
...
+        404:
+          description: Subscription with the given ID was not found
+        default:
+          $ref: '#/responses/Error'
+  /v2/subscriptions/{id}/values/current:
+    get:
+      tags:
+        - subscriptions
+      summary: Get the current values for the tags in a subscription
+      description: Get the current values for the tags in the subscription. Provides support for getting either all values, or only new values since the last request.
```
jjaco2: Returns

Provides support for getting --> Can return[? if that's correct]

Delete comma after "values"

tag/nitag.yml
```diff
@@ -0,0 +1,1814 @@
...
+        401:
+          $ref: '#/responses/Unauthorized'
+        default:
+          $ref: '#/responses/Error'
+  /v1/tags-rpc/subscriptions/heartbeat:
+    post:
+        tags:
+          - subscriptions
+        summary: Update the heartbeat timer of multiple subscriptions
+        description: Updates the heartbeat timers of multiple subscriptions. Note that this route is deprecated.
```
jjaco2: Can remove "Note that"

What does deprecated mean in the scope of this API? Asking because I've seen it mean different things among developers. Also, what do we want them to use instead?

tag/nitag.yml
```diff
@@ -0,0 +1,1814 @@
...
+        operationId: RpcUpdateSubscriptionHeartbeats
+        parameters:
+          - in: body
+            name: subscriptionIds
+            description: IDs of the subscriptions whose heartbeat timers should be updated
```
jjaco2: should be updated --> you want to update

Let me know if you think "you" feels weird in Swagger docs. We use it all the time in node documentation, for what it's worth.

tag/nitag.yml
```diff
@@ -0,0 +1,1814 @@
...
+              properties:
+                subscriptionIds:
+                  type: array
+                  items:
+                    type: string
+        responses:
+          200:
+            description: Success
+            schema:
+              description: Update subscriptions heartbeat response
```
jjaco2: Updates subscription's

tag/nitag.yml
```diff
@@ -0,0 +1,1814 @@
...
+              type: object
+              title: Update Subscriptions Heartbeat Response
```
jjaco2: Subscription's 

(I think. We're talking about one here, right?)

tag/nitag.yml
```diff
@@ -0,0 +1,1814 @@
...
+          401:
+            $ref: '#/responses/Unauthorized'
+          default:
+            $ref: '#/responses/Error'
+  /v2/refresh-subscriptions:
+    post:
+      tags:
+        - subscriptions
+      summary: Update the heartbeat timer of multiple subscriptions (SystemLink Server only)
+      description: Updates the heartbeat timers of multiple subscriptions
```
jjaco2: of --> for

tag/nitag.yml
```diff
@@ -0,0 +1,1814 @@
...
+        401:
+          $ref: '#/responses/Unauthorized'
+        default:
+          $ref: '#/responses/Error'
+  /v1/tags-rpc/subscriptions/query-values:
+    post:
+      tags:
+      - subscriptions
+      summary: Get the current values for the tags from multiple subscriptions
+      description: Get the current values for all tags in multiple subscriptions. Note that this route is deprecated.
```
jjaco2: Returns

> for all tags in multiple subscriptions

Does that mean it gets the values for all tags that happen to belong to multiple subscriptions? Or does it mean you're specifying multiple subscriptions and getting the values of all the tags therein?

If the latter, which I think it is: Returns tag values from multiple subscriptions.

tag/nitag.yml
```diff
@@ -0,0 +1,1814 @@
...
+            $ref: '#/responses/GetSubscriptionUpdatesResponse'
+        401:
+          $ref: '#/responses/Unauthorized'
+        404:
+          description: Subscription with the given ID was not found
+  /v2/query-subscription-values:
+    post:
+      tags:
+        - subscriptions
+      summary: Get the current values for the tags from multiple subscriptions (SystemLink Server only)
```
jjaco2, line 328: > (SystemLink Server only)

I could be misremembering, but did we say this differently in the Messages doc?

tag/nitag.yml
```diff
@@ -0,0 +1,1814 @@
...
+      description: Get the current values for all tags in multiple subscriptions
```
jjaco2: Should maybe be switched with Summary.

tag/nitag.yml
```diff
@@ -0,0 +1,1814 @@
...
+        404:
+          description: Subscription with the given ID was not found
+        default:
+          $ref: '#/responses/Error'
+  /v1/tags-rpc/subscriptions/delete:
+    post:
+      tags:
+        - subscriptions
+      summary: Delete multiple subscriptions
+      description: Delete one or more subscriptions based on the request body. Note that this route is deprecated.
```
jjaco2: Deletes

"based on the request body" --> "you specify"

Same note about deprecation

tag/nitag.yml
```diff
@@ -0,0 +1,1814 @@
...
+        401:
+          $ref: '#/responses/Unauthorized'
+        default:
+          $ref: '#/responses/Error'
+  /v2/selections:
+    post:
+      tags:
+        - selections
+      summary: Create a selection
+      description: Create a selection containing a collection of tags to perform subsequent operations on.
```
jjaco2: Creates a group of tags?
Selects a collection of tags?
Specifies a collection of tags?

> to perform subsequent operations on

for subsequent operations

tag/nitag.yml
```diff
@@ -0,0 +1,1814 @@
...
+          $ref: '#/responses/Unauthorized'
+        409:
+          description: Duplicate selection
+        default:
+          $ref: '#/responses/Error'
+    get:
+      tags:
+        - selections
+      summary: Get all selections
+      description: Get all selections.
```
jjaco2, line 448: Returns

Also, how long does the tag service "remember" selections? Might not affect this. Just curious.

tag/nitag.yml
```diff
@@ -0,0 +1,1814 @@
...
+        404:
+          description: Selection with the given ID not found
+        default:
+          $ref: '#/responses/Error'
+  /v2/selections/{id}:
+    put:
+      tags:
+        - selections
+      summary: Update a selection
+      description: Modify a selection
```
jjaco2: Maybe "Modifies a selection to reflect most recent value?"

Update and modify alone look like they mean different things when juxtaposed like this

tag/nitag.yml
```diff
@@ -0,0 +1,1814 @@
...
+          $ref: '#/responses/Unauthorized'
+        404:
+          description: Selection with the given ID was not found
+        default:
+          $ref: '#/responses/Error'
+    get:
+      tags:
+        - selections
+      summary: Get a selection
+      description: Get a selection's metadata.
```
jjaco2: Returns

tag/nitag.yml
```diff
@@ -0,0 +1,1814 @@
...
+          $ref: '#/responses/Unauthorized'
+        404:
+          description: Selection with the given ID was not found
+        default:
+          $ref: '#/responses/Error'
+    delete:
+      tags:
+        - selections
+      summary: Delete a selection
+      description: Delete a selection
```
jjaco2: Deletes

Does this mean it just cancels the act of selecting them? Or does it literally delete all the tags in that selection?

tag/nitag.yml
```diff
@@ -0,0 +1,1814 @@
...
+          description: Successfully deleted
+        401:
+          $ref: '#/responses/Unauthorized'
+        404:
+          description: Selection with the given ID was not found
+  /v2/selections/{id}/tags-with-values:
+    get:
+      tags:
+        - selections
+      summary: Get the tags in a selection, including their current and aggregate values
```
jjaco2, line 532: Can delete everything after the comma -- that seems like it'd belong better in the description

tag/nitag.yml
```diff
@@ -0,0 +1,1814 @@
...
+      description: Get the tags in a selection, including their current and aggregate values
```
jjaco2: Returns

tag/nitag.yml
```diff
@@ -0,0 +1,1814 @@
...
+        404:
+          description: Selection with the given ID was not found
+        default:
+          $ref: '#/responses/Error'
+  /v2/selections/{id}/tags:
+    get:
+      tags:
+        - selections
+      summary: Get the tags in a selection
+      description: Get the metadata of the tags in a selection
```
jjaco2: This seems like it means something different than the summary. Is there a difference between getting tags and getting tag metadata?

tag/nitag.yml
```diff
@@ -0,0 +1,1814 @@
...
+          $ref: '#/responses/Unauthorized'
+        404:
+          description: Selection with the given ID was not found
+        default:
+          $ref: '#/responses/Error'
+    delete:
+      tags:
+        - selections
+      summary: Delete all of the tags in a selection
+      description: Delete all of the tags in a seletion
```
jjaco2: Deletes

selection

tag/nitag.yml
```diff
@@ -0,0 +1,1814 @@
...
+        404:
+          description: Selection with the given ID was not found
+        default:
+          $ref: '#/responses/Error'
+  /v2/selections/{id}/values:
+    get:
+      tags:
+        - selections
+      summary: Get the values of the tags in a selection, excluding tag metadata
+      description: Get the current values of the tags in a given selection. The metadata for the tags will not be included.
```
jjaco2: Returns current values of tags in a selection. Does not return tag metadata.

tag/nitag.yml
```diff
@@ -0,0 +1,1814 @@
...
+        404:
+          description: Selection with the given ID was not found
+        default:
+          $ref: '#/responses/Error'
+  /v2/selections/{id}/values/current:
+    get:
+      tags:
+        - selections
+      summary: Get the current values of the tags in a selection, including timestamps
+      description: Get the current values of the tags in a selection, including timestamps
```
jjaco2: Returns current values of tags in a selection, including timestamps for each value.

tag/nitag.yml
```diff
@@ -0,0 +1,1814 @@
...
+        404:
+          description: Selection with the given ID was not found
+        default:
+          $ref: '#/responses/Error'
+  /v2/selections/{id}/values/current/value:
+    get:
+      tags:
+        - selections
+      summary: Get the current values of the tags in a selection, excluding timestamps
+      description: Get the current values of the tags in a selection, excluding timestamps
```
jjaco2: Returns

tag/nitag.yml
```diff
@@ -0,0 +1,1814 @@
...
+        404:
+          description: Selection with the given ID was not found
+        default:
+          $ref: '#/responses/Error'
+  /v2/selections/{id}/values/current/timestamp:
+    get:
+      tags:
+        - selections
+      summary: Get the timestamps of the current values of the tags in a selection
+      description: Get the timestamps of the current values of the tags in a selection
```
jjaco2: Gets the timestamps for the current values...

tag/nitag.yml
```diff
@@ -0,0 +1,1814 @@
...
+        - in: path
+          name: id
+          description: Selection ID
+          type: string
+          required: true
+      responses:
+        200:
+          description: Success
+          schema:
+            description: Array containing the paths of the tags in the selection, along with the timestamp of their current values.
```
jjaco2: Can delete period

tag/nitag.yml
```diff
@@ -0,0 +1,1814 @@
...
+              properties:
+                timestamp:
+                  description: Timestamp associated with the current value
+                  type: string
+                path:
+                  type: string
+        401:
+          $ref: '#/responses/Unauthorized'
+        404:
+          description: Selection with the given ID was not found
```
jjaco2: given --> specified

tag/nitag.yml
```diff
@@ -0,0 +1,1814 @@
...
+        default:
+          $ref: '#/responses/Error'
+  /v2/selections/{id}/values/aggregates/min:
+    get:
+      tags:
+        - selections
+      summary: Get the "min" aggregate values for the tags in a selection
+      description: Get the "min" aggregate values for the tags in a selection
```
jjaco2: Returns the minimum...

or

Returns aggregate of all minimum values for tags in a selection

tag/nitag.yml
```diff
@@ -0,0 +1,1814 @@
...
+        - in: path
+          name: id
+          description: Selection ID
+          type: string
+          required: true
+      responses:
+        200:
+          description: Success
+          schema:
+            description: Array containing the "min" values from the aggregates for each tag in the selection.
```
jjaco2: containing the --> of

Unless this array contains other values besides the min values

tag/nitag.yml
```diff
@@ -0,0 +1,1814 @@
...
+        404:
+          description: Selection with the given ID was not found
+        default:
+          $ref: '#/responses/Error'
+  /v2/selections/{id}/values/aggregates/max:
+    get:
+      tags:
+        - selections
+      summary: Get the "max" aggregate values for the tags in a selection
+      description: Get the "max" aggregate values for the tags in a selection
```
jjaco2: Same notes for "min" version

tag/nitag.yml
```diff
@@ -0,0 +1,1814 @@
...
+        - in: path
+          name: id
+          description: Selection ID
+          type: string
+          required: true
+      responses:
+        200:
+          description: Success
+          schema:
+            description: Array containing the "max" values from the aggregates for each tag in the selection.
```
jjaco2: Same commetns for "min" version

tag/nitag.yml
```diff
@@ -0,0 +1,1814 @@
...
+        404:
+          description: Selection with the given ID was not found
+        default:
+          $ref: '#/responses/Error'
+  /v2/selections/{id}/values/aggregates/mean:
+    get:
+      tags:
+        - selections
+      summary: Get the "avg" aggregate values for the tags in a selection
+      description: Get the "avg" aggregate values for the tags in a selection
```
jjaco2: Same notes as min and max versions

tag/nitag.yml
```diff
@@ -0,0 +1,1814 @@
...
+        - in: path
+          name: id
+          description: Selection ID
+          type: string
+          required: true
+      responses:
+        200:
+          description: Success
+          schema:
+            description: Array containing the "avg" values from the aggregates for each tag in the selection.
```
jjaco2: Sames notes as above

tag/nitag.yml
```diff
@@ -0,0 +1,1814 @@
...
+        404:
+          description: Selection with the given ID was not found
+        default:
+          $ref: '#/responses/Error'
+  /v2/selections/{id}/values/aggregates/count:
+    get:
+      tags:
+        - selections
+      summary: Get the "count" aggregate values for the tags in a selection
+      description: Get the "count" aggregate values for the tags in a selection
```
jjaco2: Same as others: Gets + rewording for clarity

tag/nitag.yml
```diff
@@ -0,0 +1,1814 @@
...
+        - in: path
+          name: id
+          description: Selection ID
+          type: string
+          required: true
+      responses:
+        200:
+          description: Success
+          schema:
+            description: Array containing the "count" values from the aggregates for each tag in the selection.
```
jjaco2: Same note

tag/nitag.yml
```diff
@@ -0,0 +1,1814 @@
...
+        404:
+          description: Selection with the given ID was not found
+        default:
+          $ref: '#/responses/Error'
+  /v2/selections/{id}/reset-aggregates:
+    post:
+      tags:
+        - selections
+      summary: Reset the aggregates of all tags in a selection
+      description: Reset the aggregates of all tags in a selection
```
jjaco2: Resets the aggregate values for

tag/nitag.yml
```diff
@@ -0,0 +1,1814 @@
...
+        404:
+          description: Selection with the given ID not found
+        default:
+          $ref: '#/responses/Error'
+  /v2/update-tags:
+    post:
+      tags:
+        - tags
+      summary: 'Create or update multiple tags'
+      description: Create or update multiple tags in one request.
```
jjaco2: Creates or updates

tag/nitag.yml
```diff
@@ -0,0 +1,1814 @@
...
+      operationId: CreateOrUpdateTags
+      parameters:
+        - in: body
+          name: tagsAndMerge
+          description: The tags to create or update, and a flag specifying whether keywords and properties should be merged for updates.
```
jjaco2: Tags to create or update. Also specifies whether to merge keywords and properties when updating.

tag/nitag.yml
```diff
@@ -0,0 +1,1814 @@
...
+            type: object
+            title: Tag List and Merge Flag
+            properties:
+              tags:
+                description: List of tags
+                type: array
+                items:
+                  $ref: '#/definitions/Tag'
+              merge:
+                description: 'If set, performs a merge of any specified Keywords or Properties.'
```
jjaco2: I'm probably missing something, but why is this and the next description in quotes?

tag/nitag.yml
```diff
@@ -0,0 +1,1814 @@
...
+        409:
+          description: 'Conflict: Type change not supported'
+        default:
+          $ref: '#/responses/Error'
+  /v2/tags:
+    post:
+      tags:
+        - tags
+      summary: Create or update a tag
+      description: Create or update single tag. Performs a merge of any specified Keywords or Properties. Changing a tag's type is not supported.
```
jjaco2: Creates or updates a single tag and merges any keywords or properties you specify. You cannot change a tag's data type.

tag/nitag.yml
```diff
@@ -0,0 +1,1814 @@
...
+          $ref: '#/responses/Unauthorized'
+        409:
+          description: 'Conflict: Type change not supported'
+        default:
+          $ref: '#/responses/Error'
+    get:
+      tags:
+        - tags
+      summary: Get multiple tags
+      description: Get a list of tags, filtered based on the provided query parameters.
```
jjaco2: Returns a list of tags based on query parameters you specify.

tag/nitag.yml
```diff
@@ -0,0 +1,1814 @@
...
+      operationId: GetTags
+      parameters:
+        - in: query
+          name: path
+          description: Comma-separated list of tag paths. May include glob-style wildcards
```
jjaco2, line 931: > glob-style wildcards

Great name for a poetry collection

But also, if we don't expect the novice user to know what this means, perhaps rephrase for clarity.

tag/nitag.yml
```diff
@@ -0,0 +1,1814 @@
...
+          type: string
+          required: false
+        - in: query
+          name: keywords
+          description: Comma-separated list of case-sensitive keywords.
+          type: string
+          required: false
+        - in: query
+          name: properties
+          description: Comma-separated list of name=value strings, e.g. key1=value,key2=value.
```
jjaco2: Do these lists have spaces after the commas or no? Also, idk why this line is yellow .

tag/nitag.yml
```diff
@@ -0,0 +1,1814 @@
...
+          type: string
+          required: false
+        - in: query
+          name: skip
+          description: Used for pagination. Specifies the number of entries to skip in the response list. Typically used in combination with the take parameter.
```
jjaco2: Number of entries to skip in the response list. Use with the 'take' parameter to [what do they use it for? pagination?]. 

2 questions here: 
1. Does "used for pagination" mean it helps jump around in a list of responses? I couldn't find an industry-specific definition for pagination.
2. Is there a way to clarify that take is a parameter? I used single quotes above but I'm not sure if that's misleading.

tag/nitag.yml
```diff
@@ -0,0 +1,1814 @@
...
+          type: integer
+          required: false
+        - in: query
+          name: take
+          description: Used for pagination. Specfies the number of entries to include in the response list. Typically used in combination with the skip parameter.
```
jjaco2: Same notes as above.

tag/nitag.yml
```diff
@@ -0,0 +1,1814 @@
...
+        401:
+          $ref: '#/responses/Unauthorized'
+        default:
+          $ref: '#/responses/Error'
+  /v2/tags-count:
+    get:
+      tags:
+        - tags
+      summary: Get the total number of tags
+      description: Get a count of the total number of tags.
```
jjaco2: Returns the total number of tags.

tag/nitag.yml
```diff
@@ -0,0 +1,1814 @@
...
+          description: Contains the number of tags
+        401:
+          $ref: '#/responses/Unauthorized'
+        default:
+          $ref: '#/responses/Error'
+  /v2/tags-with-values:
+    get:
+      tags:
+        - tags
+      summary: Get the current values of multiple tags
```
jjaco2, line 987: Returns current values of multiple tags.

tag/nitag.yml
```diff
@@ -0,0 +1,1814 @@
...
+          name: take
+          description: Used for pagination. Specfies the number of entries to include in the response list. Typically used in combination with the skip parameter.
+          type: integer
+          required: false
+  /v2/tags-with-values/{path}:
+    get:
+      tags:
+        - tags
+      summary: Get a tag and its current value
+      description: Get a tag and its value
```
jjaco2: What's the difference between returning a tag and returning the value of a tag? Metadata?

Also, this is another place where the description and summary seem switched.

tag/nitag.yml
```diff
@@ -0,0 +1,1814 @@
...
+        401:
+          $ref: '#/responses/Unauthorized'
+        default:
+          $ref: '#/responses/Error'
+  /v2/tags/{path}:
+    put:
+      tags:
+        - tags
+      summary: Create or replace a tag
+      description: Replace the tag at the path, or create one if one doesn''t exist. A complete resource representation must be sent (no partial updates).
```
jjaco2: Replaces the tag at the location the path specifies. If a tag does not exist where the path specifies, this route creates one. 

> A complete resource representation must be sent (no partial updates).

Should this go in a parameter description instead?

tag/nitag.yml
```diff
@@ -0,0 +1,1814 @@
...
+          description: 'No Content: Tag replaced successfully'
+        401:
+          $ref: '#/responses/Unauthorized'
+        409:
+          description: 'Conflict: Type change not supported'
+    get:
+      tags:
+        - tags
+      summary: Get a tag
+      description: Get the metadata of a single tag.
```
jjaco2: Returns the metadata for a tag.

tag/nitag.yml
```diff
@@ -0,0 +1,1814 @@
...
+            $ref: '#/definitions/Tag'
+        401:
+          $ref: '#/responses/Unauthorized'
+        default:
+          $ref: '#/responses/Error'
+    delete:
+      tags:
+        - tags
+      summary: Delete a tag
+      description: Deletes a tag
```
jjaco2, line 1105: This might not affect the description, but just for context: what happens when you delete a tag? Thinking particularly about subscriptions and selections that have already happened.

tag/nitag.yml
```diff
@@ -0,0 +1,1814 @@
...
+        404:
+          description: Not Found
+        default:
+          $ref: '#/responses/Error'
+  /v2/tags/{path}/values:
+    get:
+      tags:
+        - tags
+      summary: Get the aggregates and current value of a tag
+      description: Get the current and aggregate values for the tag at the path.
```
jjaco2: Returns

tag/nitag.yml
```diff
@@ -0,0 +1,1814 @@
...
+        404:
+          description: Tag not found
+        default:
+          $ref: '#/responses/Error'
+  /v2/tags/{path}/values/current:
+    get:
+      tags:
+        - tags
+      summary: Get the current value of a tag
+      description: Get the current value of the tag at the path.
```
jjaco2: Returns

tag/nitag.yml
```diff
@@ -0,0 +1,1814 @@
...
+          $ref: '#/responses/Unauthorized'
+        404:
+          description: Tag not found
+        default:
+          $ref: '#/responses/Error'
+    put:
+      tags:
+        - tags
+      summary: Update the current value of a tag
+      description: Update the current value of the tag at the path.
```
jjaco2: Updates

tag/nitag.yml
```diff
@@ -0,0 +1,1814 @@
...
+        409:
+          description: Conflict
+        default:
+          $ref: '#/responses/Error'
+  /v2/tags/{path}/values/current/timestamp:
+    get:
+      tags:
+        - tags
+      summary: Get the timestamp of a tag's current value
+      description: Get tag most recent timestamp.
```
jjaco2: Returns

tag/nitag.yml
```diff
@@ -0,0 +1,1814 @@
...
+        404:
+          description: Tag not found
+        default:
+          $ref: '#/responses/Error'
+  /v2/tags/{path}/values/current/value:
+    get:
+      tags:
+        - tags
+      summary: Get the current value of a tag without its timestamp
+      description: Get the current value of a tag
```
jjaco2: Returns

tag/nitag.yml
```diff
@@ -0,0 +1,1814 @@
...
+          $ref: '#/responses/Unauthorized'
+        404:
+          description: Tag with the given path was not found
+        default:
+          $ref: '#/responses/Error'
+  /v2/tags/{path}/reset-aggregates:
+    post:
+      tags:
+        - tagAggregates
+      summary: 'Reset the aggregates of a tag'
```
jjaco2: Same question about the quotes here

tag/nitag.yml
```diff
@@ -0,0 +1,1814 @@
...
+      description: Reset tag aggregates
```
jjaco2: Resets aggregate values for a tag.

tag/nitag.yml
```diff
@@ -0,0 +1,1814 @@
...
+        404:
+          description: Tag with the given path was not found
+        default:
+          $ref: '#/responses/Error'
+  /v1/tags-rpc/update-current-values:
+    post:
+      tags:
+        - tags
+      summary: Update the current values of multiple tags
+      description: Push multiple current value updates to multiple tags in one request. Note that this route is deprecated.
```
jjaco2: Updates multiple tags to their current value. This route is deprecated. [Use x instead?]

Same question as before about how we should handle deprecated routes.

tag/nitag.yml
```diff
@@ -0,0 +1,1814 @@
...
+        401:
+          $ref: '#/responses/Unauthorized'
+        default:
+          $ref: '#/responses/Error'
+  /v2/update-current-values:
+    post:
+      tags:
+        - tags
+      summary: Update the current values of multiple tags (SystemLink Server only)
+      description: Push multiple current value updates to multiple tags in one request.
```
jjaco2: Updates multiple tags to their current values.

tag/nitag.yml
```diff
@@ -0,0 +1,1814 @@
...
+        401:
+          $ref: '#/responses/Unauthorized'
+        default:
+          $ref: '#/responses/Error'
+  /v2/tags/{path}/update-values:
+    post:
+      tags:
+        - tags
+      summary: Send multiple value updates to a tag
+      description: Push a list of current value update objects to update the value of the tag at the path multiple times. This can be used to send value history in bulk.
```
jjaco2: Possible rewording, but I'm not sure if this is accurate:

Updates the current value of a tag at multiple times you specify. Use this route to send tag value history in bulk.

tag/nitag.yml
```diff
@@ -0,0 +1,1814 @@
...
+      resourceType:
+      resourceId:
+      message: "No such tag: 'foo'"
+      args: ['foo']
+  TagType:
+    type: string
+    title: Tag Data Type
+    enum: &TYPES [DOUBLE, INT, STRING, BOOLEAN, U_INT64, DATE_TIME]
+  TagWithValue:
+    description: A tag and its current value
```
jjaco2: Same question as before about tag v.s. value of tag. Does "tag" mean tag name, tag metadata, or what?

tag/nitag.yml
```diff
@@ -0,0 +1,1814 @@
...
+          value: "3.14"
+        timestamp: '2018-09-04T18:45:08Z'
+      aggregates:
+        sum: "10.5"
+        min: "0.0"
+        max: "5.0"
+        mean: "2.0"
+        count: 5
+  TagValue:
+    description: Value of a tag
```
jjaco2: What's the difference between this and current value of a tag below?

Is this one just used in operations that deal with past values?

tag/nitag.yml
```diff
@@ -0,0 +1,1814 @@
...
+    example:
+      type: DOUBLE
+      properties:
+        key1: value1
+        key2: value2
+      path: foo
+      keywords: [fooKeyword, barKeyword]
+      collectAggregates: true
+  TagUpdate:
+    description: Object containing a tag's path, and one or more updates to be applied to it
```
jjaco2: Are updates values, or separate actions?

jjaco2 (commented):
Edits mostly fall in one of these categories:

- Changing verbs to present tense in descriptions
- Ways to rephrase route names/summaries to add value or further explain stuff in the descriptions
- Clarifying technical details

Also, since there is so much reuse in this document (which is good!), I usually only commented on the first instance of an issue. Sometimes I'd comment on subsequent instances with a note like "see above." 

I don't know what the authoring process looks like on your end, but I guess be sure any changes to a phrase make it to all the places the phrase is used.

tag/nitag.yml
pvallone: This is the same thing as subscribing to a tag.

tag/nitag.yml
pvallone: For all comments like this, I only applied it to the description, as I think we wanted to keep the verb tenses as they were on the summaries. Let me know if I got that wrong.

tag/nitag.yml
pvallone: Active subscriptions are subscriptions which the subscription service is keeping up to date with respect to tag updates. Subscriptions go inactive if they haven't been used in a while, and when the service restarts.

tag/nitag.yml
pvallone: Yes, this only works for one at a time. /v2/delete-subscriptions works for multiple.

tag/nitag.yml
pvallone: This prevents subscriptions from going inactive. I updated the description: Updates the heartbeat timer for the specified subscription. If a subscription's heartbeat timer elapses, the subscription will go inactive and will no longer record updates.

tag/nitag.yml
pvallone: Yep, that's correct

tag/nitag.yml
pvallone: Deprecated means that we've replaced this route with another route, and we would prefer folks use the replacement. We're not going to add functionality to the old route, and we'll be hesitant to fix bugs in it. We want them to use /v2/refresh-subscriptions instead, and I updated the description to mention that. I did that for all of the deprecated routes.

tag/nitag.yml
pvallone: It does feel a little unusual, but I think it's fine.

tag/nitag.yml
pvallone: We're talking about the response for the update subscriptions route. The route can update multiple subscriptions. Is it still appropriate to make it possessive?

tag/nitag.yml
pvallone: It's the latter

tag/nitag.yml
pvallone: No, we phrased it the same way in that doc.

tag/nitag.yml
pvallone: I've been trying to put the (SystemLink Server Only) message in the summary so that it is clear that a route is only supported in SystemLink Server without expanding the route in Swagger UI. I changed the description to "Returns tag values from multiple subscriptions" - let me know if you think we should still swap the summary.

tag/nitag.yml
pvallone: A selection is a resource that the tag service manages and introduces, but I like "Selects a collection of tags for subsequent operations" so long as the summary remains "Create a selection", so that folks know what we're calling that resource.

tag/nitag.yml
pvallone: The tag service remembers selections for up to fifteen minutes. I added a note of this to the description for creating a selection.

tag/nitag.yml
pvallone: I went with "Modifies a selection's metadata", as we're not doing anything related to tag values here.

tag/nitag.yml
pvallone: It just cancels the act of selecting them. There's a different route for deleting all of the tags in a selection: DELETE /v2/selections/{id}/tags.

tag/nitag.yml
pvallone: These summary strings are best viewed in the context of Swagger UI. We have another route which shows up right next to this one in Swagger UI for getting the tags in a selection *without* their current and aggregate values, and so we need to have something to distinguish the two routes in the summary. Feel free to drop by my desk, or copy/paste the document in editor.swagger.io if you want to see what I'm talking about.

tag/nitag.yml
pvallone: I've been trying to refer to "getting a tag" as just getting the tag's metadata (path, keywords, properties), and then adding more to the summary and description if we're also getting things like current and aggregate values. We have a lot of routes which just get metadata, and also a lot of routes which get metadata in addition to current and aggregate values. 

Do you think it would be good to modify the summaries of the routes which just get tag metadata to "Get a tag's metadata"?

tag/nitag.yml
pvallone: I went with the first one, as it's the minimum value from each "aggregate" object.

tag/nitag.yml
pvallone: This one didn't need to be in quotes. I tried to remove the quotes from descriptions which don't need them. Some strings contain restricted words, like "Created" or "Conflicts", and so they need to be quoted. I'll clean the rest up.

tag/nitag.yml
pvallone: I think glob is a standard enough term that folks who are going to be using a REST API will understand (https://en.wikipedia.org/wiki/Glob_(programming)).

tag/nitag.yml
pvallone: I think they can, but you have to URL-encode them, and I didn't want folks to have to worry about that in the examples. Hm, it isn't yellow on my screen.

tag/nitag.yml
pvallone: They use it for pagination, yes. I removed the first sentence and reworded it like this: Number of entries to skip in the response list. Typically used in combination with the take parameter to implement pagination.

1. Yes. If your query returned too many tags to include in one response, you use skip and take to get the rest of your data. I think skip and take are pretty intuitive for clients of REST APIs, so I'm not too worried about it.
2. Take is specified under the "parameters" section, so it appears in Swagger UI's "parameters" section. I think that should be sufficient.

tag/nitag.yml
pvallone: I added this in a new description line, as I think that's what you meant for me to modify.

tag/nitag.yml
pvallone: See my other response for the first question. And yeah, I swapped the summary and description.

tag/nitag.yml
pvallone: Yep! Added it to the parameter description.

tag/nitag.yml
pvallone: I believe the subscription notifies you about tags that have been deleted or couldn't be found. I think the tag would just disappear from the selection.

tag/nitag.yml
pvallone: In the context of this doc, I've tried to just have "tag" mean tag metadata, so path, properties, keywords, and type. If current and aggregate values are included, I try to call them out separately.

tag/nitag.yml
pvallone: This one just contains the value and the type, while TagCurrentValue also contains a timestamp. This one is just used in contexts where a route only returns a value and a type, as some of the routes provide support for projecting out timestamps. Thanks for pointing this out though, as I don't like the name "TagCurrentValue". I changed it to "TimestampedTagValue", as that makes it more clear.

tag/nitag.yml
pvallone: Updates are values - a list of tag values with timestamps which should be applied to a tag.

tag/nitag.yml
jjaco2: I had looked at this in Swagger UI, but hadn't thought of that risk of repetition. I agree with you now! Keep it that way.

tag/nitag.yml
pvallone: That's accurate

tag/nitag.yml
```diff
@@ -0,0 +1,1816 @@
+swagger: '2.0'
+info:
+  description: Publish and manage timestamped key-value-pair data.
+  version: '1'
```
spanglerco: Should this be 2?

tag/nitag.yml
```diff
@@ -0,0 +1,1816 @@
...
+      summary: Create a subscription
+      description: Creates a subscription for a list of tags to get notifications when the tags change.
+      operationId: CreateSubscription
+      parameters:
+        - in: body
+          name: subscription
+          description: Subscription to create
+          schema:
+            $ref: '#/definitions/Subscription'
+          required: false
```
spanglerco: This should be required (and appears to be for SystemLink Server).

tag/nitag.yml
```diff
@@ -0,0 +1,1816 @@
...
+                type: string
+              nonexistentTags:
+                description: List of tags which did not exist
+                type: array
+                items:
+                  type: string
+              alias:
+                type: string
+                description: Name of the subscription
+        400:
```
spanglerco: Do we need to document the 400 explicitly? SL Server would include an Error response in a 400, which is covered by the default case.

tag/nitag.yml
```diff
@@ -0,0 +1,1816 @@
...
+          type: string
+          required: true
+      responses:
+        200:
+          schema:
+            $ref: '#/definitions/Subscription'
+          description: Success
+        401:
+          $ref: '#/responses/Unauthorized'
+        404:
```
spanglerco: Ditto on SL Server returning an error response here. Though I could see it being useful to document 404s? Though I guess it's implied that any URL with an ID in it could return a 404 if the ID doesn't match anything.

Also applies to other routes.

tag/nitag.yml
```diff
@@ -0,0 +1,1816 @@
...
+      parameters:
+        - in: path
+          name: id
+          description: Subscription ID
+          type: string
+          required: true
+      responses:
+        200:
+          schema:
+            $ref: '#/responses/GetSubscriptionUpdatesResponse'
```
spanglerco: This is a response ref, so it needs to go up a level instead of under the schema section.

tag/nitag.yml
```diff
@@ -0,0 +1,1816 @@
...
+          - subscriptions
+        summary: Update the heartbeat timer of multiple subscriptions
+        description: Updates the heartbeat timers of multiple subscriptions. This route is deprecated in favor of /v2/refresh-subscriptions.
+        operationId: RpcUpdateSubscriptionHeartbeats
+        deprecated: true
+        parameters:
+          - in: body
+            name: subscriptionIds
+            description: IDs of the subscriptions whose heartbeat timers you want to update
+            required: false
```
spanglerco: Ditto on should be required. For SL Server, the service throws an exception when it's not given (oops). subscriptionIds should probably be marked required as well.

I believe this pretty much applies to every put and post, I won't comment on each.

tag/nitag.yml
```diff
@@ -0,0 +1,1816 @@
...
+            properties:
+              subscriptionIds:
+                type: array
+                items:
+                  type: string
+      responses:
+        200:
+          description: Success
+          schema:
+            $ref: '#/responses/GetSubscriptionUpdatesResponse'
```
spanglerco: Ditto on responses ref moving out of the schema.

tag/nitag.yml
```diff
@@ -0,0 +1,1816 @@
...
```
spanglerco: Another response ref to move up.

tag/nitag.yml
```diff
@@ -0,0 +1,1816 @@
...
+      responses:
+        200:
+          description: Success
+          schema:
+            $ref: '#/definitions/Selection'
+        400:
+          description: Bad Request
+        401:
+          $ref: '#/responses/Unauthorized'
+        409:
```
spanglerco: Kind of like 404, SL Server returns an error response in this case.

tag/nitag.yml
```diff
@@ -0,0 +1,1816 @@
...
+        - in: path
+          name: id
+          description: Selection ID
+          type: string
+          required: true
+      responses:
+        200:
+          description: Success
+          schema:
+            $ref: '#/responses/GetTagsWithValuesResponse'
```
spanglerco: Ditto on response ref under schema.

tag/nitag.yml
```diff
@@ -0,0 +1,1816 @@
...
+            type: array
+            title: Tag Paths With Timestamps
+            items:
+              type: object
+              title: Tag Path and Timestamp
+              description: A tag path and the timestamp of its current value
+              properties:
+                timestamp:
+                  description: Timestamp associated with the current value
+                  type: string
```
spanglerco, line 706: For timestamps, it could be useful to give an example so it's easy to tell the expected format. For test monitor, we've documented our timestamps via `format: iso-date-time` which isn't a standard format, but the format property is extensible.

tag/nitag.yml
```diff
@@ -0,0 +1,1816 @@
...
+            type: array
+            items:
+              type: object
+              title: Tag Paths and Count Values
+              properties:
+                path:
+                  type: string
+                min:
+                  description: Minimum value from the tag's aggregates
+                  type: number
```
spanglerco: Throughout this API, tag aggregates changed from v1 to v2. In v2, aggregates are all strings (except count). In v1, they are doubles (except count is an integer). I don't know if SL Cloud has that change in their v2, but I would think so.

Applies to the other selection aggregate routes below.

tag/nitag.yml
```diff
@@ -0,0 +1,1816 @@
...
+            type: array
+            items:
+              type: object
+              title: Tag Paths and Count Values
+              properties:
+                path:
+                  type: string
+                count:
+                  description: Count value from the tag's aggregates
+                  type: number
```
spanglerco: integer would be more specific.

spanglerco (changes requested):
I'm about halfway through the review, submitting the comments I have so far. Note that I'm not reviewing comments or API accuracy.

tag/nitag.yml
```diff
@@ -0,0 +1,1816 @@
...
+          schema:
+            type: object
+            title: Tag List and Merge Flag
+            properties:
+              tags:
+                description: List of tags
+                type: array
+                items:
+                  $ref: '#/definitions/Tag'
+              merge:
```
spanglerco, line 884: We should either provide a default for merge or mark it required.

tag/nitag.yml
```diff
@@ -0,0 +1,1816 @@
...
+      operationId: CreateOrUpdateTags
+      parameters:
+        - in: body
+          name: tagsAndMerge
+          description: Tags to create or update. Also specifies whether to merge keywords and properties when updating.
+          schema:
+            type: object
+            title: Tag List and Merge Flag
+            properties:
+              tags:
```
spanglerco, line 879: The tags array should probably be marked required.

tag/nitag.yml
```diff
@@ -0,0 +1,1816 @@
...
+  /v2/tags:
+    post:
+      tags:
+        - tags
+      summary: Create or update a tag
+      description: Creates or update a single tag and merges any keywords or properties you specify. You cannot change a tag's data type.
+      operationId: CreateTag
+      parameters:
+        - in: body
+          name: tags
```
spanglerco: Should this be singular tag since it's not an array?

tag/nitag.yml
```diff
@@ -0,0 +1,1816 @@
...
+              tags:
+                description: List of tags
+                type: array
+                items:
+                  $ref: '#/definitions/Tag'
+              merge:
+                description: If set, performs a merge of any specified Keywords or Properties.
+                type: boolean
+          required: true
+      responses:
```
spanglerco: The SL Server version of this API can return a 200 for a partial success error response. So if we're able to create 2 out of 4 tags, it'll be a 200 instead of 201, and the body will contain the 2 errors for the failed tags.

tag/nitag.yml
```diff
@@ -0,0 +1,1816 @@
...
+          type: string
+          required: false
+        - in: query
+          name: keywords
+          description: Comma-separated list of case-sensitive keywords.
+          type: string
+          required: false
+        - in: query
+          name: properties
+          description: Comma-separated list of name=value strings, e.g. key1=value,key2=value.
```
spanglerco: Instead of having the example in the description, put it in as an actual example on the parameter. I think all of these string parameters could benefit from an example since they're non-trivial.

tag/nitag.yml
```diff
@@ -0,0 +1,1816 @@
...
+          type: string
+          required: false
+        - in: query
+          name: skip
+          description: Number of entries to skip in the response list. Typically used in combination with the take parameter to implement pagination.
+          type: integer
+          required: false
```
spanglerco, line 951: Should we document defaults for skip and take?

tag/nitag.yml
```diff
@@ -0,0 +1,1816 @@
...
+        - in: query
+          name: take
+          description: Number of entries to include in the response list. Typically used in combination with the skip parameter to implement pagination.
+          type: integer
+      responses:
+        200:
+          schema:
+            $ref: '#/responses/GetTagsResponse'
```
spanglerco: Another responses inside schema that needs to move up.

tag/nitag.yml
```diff
@@ -0,0 +1,1816 @@
...
+    get:
+      tags:
+        - tags
+      summary: Get the total number of tags
+      description: Returns the total number of tags.
+      operationId: GetTagCount
+      parameters: []
+      responses:
+        200:
+          description: Contains the number of tags
```
spanglerco, line 975: This and /v2/tags/{path}/values/current/timestamp have a strange return value. They appear to be scalars that actually go through the JSON serializer. So the content type is returned as application/json, but the data is just a number or a quoted string in the case of timestamp. That seems bad, and not all JSON deserializers can handle that (e.g. LabVIEW). I would think these should return text/plain instead, but that would be a separate CAR.

If nothing else, we should provide an example response so it's less confusing. Swagger UI seems a bit confused by it since it doesn't show the quotes for timestamp, but you can do this:

```yaml
200:
  schema:
    type: integer
    example: 2
```

tag/nitag.yml
```diff
@@ -0,0 +1,1816 @@
...
+  /v2/tags-with-values:
+    get:
+      tags:
+        - tags
+      summary: Get the current values of multiple tags
+      description: Returns current values of multiple tags.
+      responses:
+        200:
+          schema:
+            $ref: '#/responses/GetTagsWithValuesResponse'
```
spanglerco: Another response ref under a schema.

tag/nitag.yml
```diff
@@ -0,0 +1,1816 @@
...
+          description: Conflict
+        default:
+          $ref: '#/responses/Error'
+      operationId: GetTagsWithValues
+      parameters:
+        - in: query
+          name: path
+          description: Comma-separated list of tag paths. May include glob-style wildcards.
+          type: string
+          required: false
```
spanglerco: Ditto on providing examples for these string parameters and defaults for skip and take if applicable.

tag/nitag.yml
```diff
@@ -0,0 +1,1816 @@
...
+          description: Not Found
+        default:
+          $ref: '#/responses/Error'
+  /v2/tags/{path}/values:
+    get:
+      tags:
+        - tags
+      summary: Get the aggregates and current value of a tag
+      description: Returns the current aggregate values for the tag at the path.
+      operationId: get-tag-values
```
spanglerco: This and the next operationId being hyphenated are inconsistent with the others.

tag/nitag.yml
```diff
@@ -0,0 +1,1816 @@
...
+        - tags
+      summary: Update the current values of multiple tags
+      description: Updates multiple tags to their current value. This route is deprecated in favor of /v2/update-current-values.
+      operationId: RpcUpdateTagCurrentValues
+      deprecated: true
+      parameters:
+        - in: body
+          name: updates
+          description: Updates
+          schema:
```
spanglerco, line 1364: Should the body be marked required? Ditto for /v2/update-current-values below.

tag/nitag.yml
```diff
@@ -0,0 +1,1816 @@
...
+        401:
+          $ref: '#/responses/Unauthorized'
+        404:
+          description: Tag not found
+        default:
+          $ref: '#/responses/Error'
+  /v2/tags/{path}/values/aggregates/min:
+    get:
+      tags:
+        - tagAggregates
```
spanglerco: Tags can have spaces. How about "tag aggregates"?

tag/nitag.yml
```diff
@@ -0,0 +1,1816 @@
...
+        - in: path
+          name: path
+          description: Tag path
+          type: string
+          required: true
+      responses:
+        200:
+          description: Minimum aggregate value
+          schema:
+            type: string
```
spanglerco, line 1251: Here's another case where v1 vs. v2 is different. In v1, min, max, and mean will return a number instead of a string. Also, this is another case of the weird JSON-but-not-really.

tag/nitag.yml
```diff
@@ -0,0 +1,1816 @@
...
+      name: Tag.NoSuchTag
+      code: -251909
+      resourceType:
+      resourceId:
+      message: "No such tag: 'foo'"
+      args: ['foo']
+  TagType:
+    type: string
+    title: Tag Data Type
+    enum: &TYPES [DOUBLE, INT, STRING, BOOLEAN, U_INT64, DATE_TIME]
```
spanglerco: I think a $ref should work for TagType rather than using a YAML anchor.

tag/nitag.yml
```diff
@@ -0,0 +1,1816 @@
...
+  TagValue:
+    description: Value of a tag
+    type: object
+    title: Tag Value
+    properties:
+      value:
+        type: string
+      type:
+        type: string
+        enum: *TYPES
```
spanglerco: I believe both value and type could be marked as required.

tag/nitag.yml
```diff
@@ -0,0 +1,1816 @@
...
+    description: Current value of a tag
+    type: object
+    title: Tag Current Value
+    properties:
+      value:
+        $ref: '#/definitions/TagValue'
+      timestamp:
+        description: Timestamp associated with the value
+        type: string
+        format: date-time
```
spanglerco: The Swagger date-time format is RFC 3339, while we use ISO 8601 format. In our other documents, we've been using the custom `format: iso-date-time`.

tag/nitag.yml
```diff
@@ -0,0 +1,1816 @@
...
+        mean: "2.0"
+        count: 5
+  TagAggregates:
+    description: Aggregate values of a tag
+    type: object
+    title: Tag Aggregate Value
+    properties:
+      sum:
+        description: Sum aggregate value
+        type: number
```
spanglerco: In v2, I believe these are all strings except count.

tag/nitag.yml
```diff
@@ -0,0 +1,1816 @@
...
+    title: Tag
+    properties:
+      type:
+        description: Type
+        type: string
+        enum: *TYPES
+      properties:
+        type: object
+      path:
+        type: string
```
spanglerco: Path should probably be marked required, not sure if maybe type as well.

tag/nitag.yml
```diff
@@ -0,0 +1,1816 @@
...
+    description: Object containing a tag's path, and one or more updates to be applied to it
+    type: object
+    title: Tag Update
+    properties:
+      path:
+        type: string
+      updates:
+        type: array
+        items:
+          $ref: '$/definitions/TimestampedTagValue'
```
spanglerco: Probably mark both of these properties as required.

tag/nitag.yml
```diff
@@ -0,0 +1,1816 @@
...
+    properties:
+      value:
+        description: Updated tag value
+        type: string
+      type:
+        type: string
+        enum: *TYPES
+      timestamp:
+        description: Updated timestamp
+        type: string
```
spanglerco: Consider adding `format: iso-date-time`.

tag/nitag.yml
```diff
@@ -0,0 +1,1816 @@
...
+        description: Whether to notify only about updates
+        type: boolean
+      tags:
+        description: List of subscribed tag paths
+        type: array
+        items:
+          type: string
+      alias:
+        description: Name of the subscription
+        type: string
```
spanglerco, line 1725: Should all three of these be required? Or maybe updatesOnly would be optional (in which case we should provide a default).

tag/nitag.yml
```diff
@@ -0,0 +1,1816 @@
...
+      - searchPaths
+    properties:
+      searchPaths:
+        type: array
+        items:
+          type: string
+      inactivityTimeout:
+        description: Inactivity timeout in seconds
+        type: integer
+        maximum: 900
```
spanglerco, line 1744: Any default to document?

spanglerco (dismissed):

tag/nitag.yml
pvallone: Yeah, it makes sense for us to establish a convention that this be the latest version of the API.

tag/nitag.yml
pvallone: Probably not, falling back to the default error response is probably good enough.

tag/nitag.yml
pvallone: I think I'll remove 404s for the same reason that we removed the 400s.

tag/nitag.yml
pvallone: Added an example

tag/nitag.yml
pvallone: Cloud has not made that change. I filed them a CAR. We'll document them as strings.

tag/nitag.yml
pvallone: Marking it as false by default

tag/nitag.yml
pvallone: Marked the object they're contained in as required

tag/nitag.yml
pvallone: Yep!

tag/nitag.yml
pvallone: I try to forget about the partial success stuff.. I documented the 200 case, and just reused the error response schema, as that should still be accurate.

tag/nitag.yml
pvallone: Did so, that's much cleaner

tag/nitag.yml
pvallone: The default for skip is zero for both Cloud and Server, so I will document that. I'm not sure what Cloud's default take is, but ours is 100000. Cloud's API should be identical to ours, so if 100000 is inaccurate, I'm just going to treat that as a CAR to them.

tag/nitag.yml
pvallone: I'll make a note to CAR this next week, when I have more time to investigate. I provided an example response.

tag/nitag.yml
pvallone: Did so

tag/nitag.yml
pvallone: Yes

tag/nitag.yml
pvallone: Woops, that's left over from Cloud. Thought I got them all.

tag/nitag.yml
pvallone: Oh nice, I like that much more.

tag/nitag.yml
pvallone: Yep, switched to a $ref

tag/nitag.yml
pvallone: Switched to iso-date-time.

tag/nitag.yml
pvallone: Changed them

tag/nitag.yml
pvallone: Alias doesn't seem to be required, but updatesOnly defaults to false, and the paths are required.

tag/nitag.yml
pvallone: Yep, 900 seconds.

tag/nitag.yml
pvallone: Removed the 409s

tag/nitag.yml
pvallone: I'll address this in another PR, as this review is already too large. The v1 route for querying subscription values needs to return numeric aggregates, instead of strings.

tag/nitag.yml
harmish-modi: Subscriptions'

tag/nitag.yml
```diff
@@ -0,0 +1,1749 @@
...
+                items:
+                  type: string
+      responses:
+        200:
+          description: Success
+          schema:
+            type: object
+            title: Update Subscriptions Heartbeat Response
+            properties:
+              invalidSubscriptionIds:
```
harmish-modi, line 288: Ideally, we should make PartialSuccessResposne for the partial successes. However, this document has to resemble the existing implementations of the APIs, so no change is required, but making this response to PartialSuccessResponse should be tracked by some CAR or user story.

tag/nitag.yml
```diff
@@ -0,0 +1,1749 @@
...
+        default:
+          $ref: '#/responses/Error'
+  /v2/selections/{id}/tags:
+    get:
+      tags:
+        - selections
+      summary: Get the tags in a selection
+      description: Returns the metadata of the tags in a selection
+      operationId: GetSelectionTags
+      parameters:
```
harmish-modi, line 555: Isn't this route redundant after "GET /v2/selections/{id}"? If yes, then again this should be tracked by CAR.

pvallone: No, as GET /v2/selections/{id} gets the selection metadata, while this route gets the tags in the selection.

tag/nitag.yml
```diff
@@ -0,0 +1,1749 @@
...
+          required: true
+      responses:
+        200:
+          description: Success
+          schema:
+            description: Array of the "min" values from the aggregates for each tag in the selection.
+            type: array
+            items:
+              type: object
+              title: Tag Paths and Count Values
```
harmish-modi: Tag Paths and MinValues

tag/nitag.yml
pvallone: I filed 712984

tag/nitag.yml
pvallone: Went with "Tag Path with Min Aggregate" for consistency with the related routes.

tag/nitag.yml
```diff
@@ -0,0 +1,1749 @@
...
+      tags:
+        - tag aggregates
+      summary: Get the value of a tag's "avg" aggregate
+      description: Get the value of a tag's "avg" aggregate
+      operationId: GetTagMeanAggregate
+      responses:
+        200:
+          description: Mean aggregate value
+          schema:
+            type: string
```
harmish-modi: number.

tag/nitag.yml
```diff
@@ -0,0 +1,1749 @@
...
+        - in: path
+          name: path
+          description: Tag path
+          type: string
+          required: true
+      responses:
+        200:
+          description: Count aggregate value
+          schema:
+            type: string
```
harmish-modi: integer

tag/nitag.yml
```diff
@@ -0,0 +1,1749 @@
...
+      description: Resets aggregate values for a tag.
+      operationId: ResetTagAggregates
+      parameters:
+        - in: path
+          name: path
+          description: Tag path
+          type: string
+          required: true
+      responses:
+        200:
```
harmish-modi, line 1344: 204.

tag/nitag.yml
```diff
@@ -0,0 +1,1749 @@
...
+        - in: body
+          name: updates
+          description: Updates
+          schema:
+            type: array
+            items:
+              $ref: '#/definitions/TagUpdate'
+          required: true
+      responses:
+        200:
```
harmish-modi, line 1370: 202.

tag/nitag.yml
```diff
@@ -0,0 +1,1749 @@
...
```
harmish-modi: 202

harmish-modi: 202

tag/nitag.yml
```diff
@@ -0,0 +1,1749 @@
...
+          $ref: '#/responses/Error'
+  /v2/tags/{path}/update-values:
+    post:
+      tags:
+        - tags
+      summary: Send multiple value updates to a tag
+      description: Updates the current value of a tag at multiple times you specify. Use this route to send tag value history in bulk.
+      operationId: update-tag-values
+      responses:
+        200:
```
harmish-modi, line 1413: 202

tag/nitag.yml
pvallone: This is a string in v2.

tag/nitag.yml
```diff
@@ -0,0 +1,1749 @@
...
+      description: Updates the heartbeat timer for the specified subscription. If a subscription's heartbeat timer elapses, the subscription will go inactive and will no longer record updates.
+      operationId: UpdateSubscriptionHeartbeat
+      parameters:
+        - in: path
+          name: id
+          description: Subscription ID
+          type: string
+          required: true
+      responses:
+        200:
```
harmish-modi, line 197: 204

tag/nitag.yml
```diff
@@ -0,0 +1,1749 @@
...
+          type: string
+          required: true
+        - in: body
+          name: tagValue
+          description: Timestamped value
+          schema:
+            $ref: '#/definitions/TimestampedTagValue'
+          required: true
+      responses:
+        200:
```
harmish-modi, line 1180: 202

tag/nitag.yml
pvallone: My mistake, it's numeric.

tag/nitag.yml
pvallone: Cloud returns a 200, but server returns a 204. For all of these cases, I'll file a CAR later. I'll document what both return.

tag/nitag.yml
```diff
@@ -0,0 +1,1749 @@
...
+        max: "5.0"
+        avg: 2.0
+        count: 5
+  TagAggregates:
+    description: Aggregate values of a tag
+    type: object
+    title: Tag Aggregate Value
+    properties:
+      sum:
+        description: Sum aggregate value
```
harmish-modi: Do we support this?

pvallone: Oh, nice catch. I think we used to, but then we removed it. I removed references to it.

tag/nitag.yml
jjaco2: Sounds good. I just remember at one point someone had mentioned that some users for whom REST APIs are most useful wouldn't even be familiar with them at all before this. But looking into this more, "glob" applies to a broad enough domain that it wouldn't be wrong to assume they know.

tag/nitag.yml
jjaco2: Okay. Given that what makes up tag metadata is just those 4 things, I think this is fine.

tag/nitag.yml
jjaco2: What do you think about "and one or more values to update the tag with at times you specify?" It's longer, but clarifies the level of control the user has here.

jjaco2 (dismissed):
Replied to a few comments. Looks good! I have one rewording suggestion I added, but it's not critical.

tag/nitag.yml
pvallone: I like that a lot more, thanks!

harmish-modi (approved):

PR #5, opened 2018-09-07
Moving the path parameters into operation structure

tschmittni:
Merged the path parameters with all the other parameters
in the operation structure.

Our current tooling in SystemLink Cloud does not support parameters
on the path level.

### DIFF
--- a/message/nimessage.yml
+++ b/message/nimessage.yml
@@ -68,14 +68,14 @@ paths:
         default:
           $ref: '#/responses/Error'
   /v1/sessions/{token}:
-    parameters:
-      - $ref: '#/parameters/Token'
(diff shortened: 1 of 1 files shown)


pvallone (approved):
What custom tooling are you folks using around Swagger, just so I know for the future?

spanglerco (approved):

tschmittni:
Our CLI is using https://github.com/go-openapi/

I have added you to a pull request which enables basic auth so that the CLI can target the SystemLink Server: https://github.com/ni-kismet/helium-cli/pull/16

pvallone:
Ah alright. I approved that pull request.

PR #6, opened 2018-09-07
Remove unused definitions

tschmittni:
Just removing unused definitions to fix some warnings.


pvallone (changes requested):
These definitions need to stay, even though they aren't used. Swagger doesn't provide a way to document a WebSocket API, so all we can do is add definitions for the available WebSocket actions.

tschmittni:
I wouldn't add the WebSocket-related parts at all. Swagger doesn't support WebSockets at the moment which makes this rather useless.

pvallone:
We document how to open a WebSocket connection in Swagger, and reference the action objects in the description for that route. I agree that they're useless in that you can't do WebSocket stuff in Swagger UI, or in generated clients, but you can at least view the schema for the actions, and our thinking was that that was better than nothing.

spanglerco:
We also intend to support websockets in our codegen for services.

tschmittni:
alright, sounds good.

PR #7, opened 2018-09-07
Initial commit of the File Swagger API

BKnight760:
This is the existing YML that has already been reviewed and is in use by the File service.

This pull request adds a Swagger document for the File service. This will be the master copy of the File service's Swagger document, for use in SystemLink Server. This document is used as the source for the webservice interface which is implemented by the File service.

This pull request gives higher visibility to the File HTTP API, and ensures that the implentation matches the fully documented options for the webservice.

New automated python tests were added when the File service was switched to using the Swagger code-generation for the webservice interface.  Existing automated tests have been run against the swagger interface, and all tests pass.

### DIFF
--- a/file/nifile.yml
+++ b/file/nifile.yml
@@ -191,7 +191,7 @@ definitions:
           - LESS_THAN_OR_EQUAL
           - NOT_EQUAL
       value:
-        description: The value of the field to query for
+        description: The value of the field to search for
         type: string
         format: iso-date-time
         example: '2018-05-14T00:00:00.000Z'
@@ -216,7 +216,7 @@ definitions:
           - LESS_THAN_OR_EQUAL
           - NOT_EQUAL
       value:
-        description: The value of the field to query for
+        description: The value of the field to search for
         type: integer
         example: 5000
     required:
@@ -230,7 +230,7 @@ definitions:
     type: object
     properties:
       key:
-        description: The name of the property to query against
+        description: The name of the property to search against
         type: string
         example: Name
       operation:
@@ -242,7 +242,7 @@ definitions:
           - NOT_CONTAINS
           - NOT_EQUAL
       value:
-        description: The value of the property to query for
+        description: The value of the property to search for
         type: string
         example: myfile.txt
     required:
@@ -264,7 +264,7 @@ definitions:
           - EQUAL
           - NOT_EQUAL
       ids:
-        description: The array of ids to query for
+        description: The array of ids to search for
         type: array
         items:
           type: string
@@ -283,7 +283,7 @@ definitions:
           - EQUAL
           - NOT_EQUAL
       value:
-        description: The value of the field to query for
+        description: The value of the field to search for
         type: string
         example: txt
     required:
@@ -549,7 +549,7 @@ paths:
           default: false
         - in: query
           name: id
-          description: Comma-separated list of file IDs to restrict to in the results
+          description: Comma-separated list of file IDs to search by
           type: string
           x-example: '5afb2ce3741fe11d88838cc9,4afb2ce3741fe11d88838cc9'
       responses:
@@ -583,15 +583,11 @@ paths:
       tags: [files]
       summary: Download file
       description: >-
-        Downloads a file from the Skyline File service. The file is returned in
-        a single HTTP response. The file's name metadata property is appended
-        to the response header so that the file is downloaded with the
-        appropriate file name.When the the inline query parameter is not
-        specified or is set to false, the Content-Disposition header is set to
-        'attachment' and the MIME type is set to binary so that the browser
-        handles the response as a file download.  When the inline query
-        parameter is set to true, the Content-Disposition header and MIME type
-        remain unset, and the file contents must be handled by the caller.
+        Downloads a file from the SystemLink File service in
+        a single HTTP response.
+        
+        Use the inline parameter in the query string to control the download
+        behavior.
       operationId: ReceiveFile
       x-ni-operation: downloadData
       x-ni-privilege: DownloadFiles
@@ -600,7 +596,15 @@ paths:
       parameters:
         - in: query
           name: inline
-          description: Whether to return the file data inline rather than as an attachment.
+          description: >-
+            Whether to return the file data inline rather than as an attachment.
+            When the inline query parameter is true, the file contents return
+            directly for the caller to handle. The Content-Disposition header
+            and MIME type remain unset.
+            When the the inline query parameter is not specified or is
+            false, the file contents are handled as a download. The
+            Content-Disposition header is set to 'attachment' and
+            the MIME type is set to binary.
           type: boolean
           default: false
       produces:
--- a/file/nifile.yml
+++ b/file/nifile.yml
@@ -1,0 +1,816 @@
+swagger: '2.0'
+info:
+  version: '1'
+  title: Skyline File Service
+  description: Upload and download files from a SystemLink server
+  contact:
+    name: National Instruments
+    url: https://www.ni.com/systemlink
+    email: support@ni.com
+basePath: /nifile
+consumes: [application/json]
+produces: [application/json]
+
+securityDefinitions:
+  basicAuth:
+    type: basic
+  cookieAuth:
+    type: apiKey
+    in: header
+    name: Cookie
+
+security:
+  - basicAuth: []
+  - cookieAuth: []
+x-ni-routing-key: Skyline.FileIngestion
+definitions:
+  Link:
+    description: A hyperlink for a resource or action on a resource
+    type: object
+    properties:
+      href:
+        type: string
+        description: URI of the link
+        example: /nifile/v1/service-groups
+    required: [href]
+  Error:
+    description: Contains error information
+    type: object
+    properties:
+      name:
+        description: String error code
+        type: string
+      code:
+        description: Numeric error code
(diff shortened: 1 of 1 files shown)


spanglerco (dismissed):

file/nifile.yml
```diff
@@ -0,0 +1,816 @@
...
+    properties:
+      name:
+        type: string
+        description: The service group's name
+        example: Default
+      _links:
+        description: >-
+          The links that apply to a service group:
+
+          - deleteFiles: Link to delete multiple files from the service group using a POST
```
jjaco2: Here and in all of these: could we delete "a" before the method to make it easier to read?

file/nifile.yml
```diff
@@ -0,0 +1,816 @@
...
+          query:
+            href: /nifile/v1/service-groups/Default/query-files
+          searchFiles:
+            href: /nifile/v1/service-groups/Default/files
+          self:
+            href: /nifile/v1/service-groups/Default
+          upload:
+            href: /nifile/v1/service-groups/Default/upload-files
+  DateQuery:
+    description: A query for an date and time field
```
jjaco2: a date

file/nifile.yml
```diff
@@ -0,0 +1,816 @@
...
+        type: string
+        enum:
+          - EQUAL
+          - GREATER_THAN
+          - GREATER_THAN_OR_EQUAL
+          - LESS_THAN
+          - LESS_THAN_OR_EQUAL
+          - NOT_EQUAL
+      value:
+        description: The value of the field to query for
```
jjaco2: Here and below: query --> search?

file/nifile.yml
```diff
@@ -0,0 +1,816 @@
...
+              href: /nifile/v1/service-groups/Default/files
+            upload:
+              href: /nifile/v1/service-groups/Default/upload-files
+        availableFiles:
+          description: The list of files returned by the query
+          type: array
+          items:
+            $ref: '#/definitions/FileMetadata'
+        totalCount:
+          description: The total number of files that match the query regardless of skip and take
```
jjaco2: I think I asked this in one of Phil's reviews, but I haven't seen if he replied to it: is there a way to make it clear that skip and take are parameters? If not, maybe add "values" to the end of the sentence.

file/nifile.yml
```diff
@@ -0,0 +1,816 @@
...
+    get:
+      tags: [files]
+      summary: List available files
+      description: >-
+        Lists available files on the Skyline File service.
+
+        The skip and take parameters can be used return paged responses. The
+        user can skip an unsigned integer number of entries, and take an
+        unsigned integer number of entries. A list of 100 files with a skip
+        value of 50 and a take value of 25 will return entries 51 through 75.
```
jjaco2: Use the skip and take parameters together to return paged responses. Specify a number of entries to skip and a number of entries to return after skipping. For example, a list of 100 files with a skip value of 50 and a take value of 25 will return entries 51 through 75.

jjaco2: I'm also wondering if any of this belongs in parameter descriptions instead of the route description. Just something to think about.

file/nifile.yml
```diff
@@ -0,0 +1,816 @@
...
+
+        The orderBy and orderByDescending fields are intended to manage sorting
+        the list by metadata objects. The value of the orderBy field should be
+        the name of a metadata key. The elements in the list are sorted
+        ascending by default. If the orderByDescending parameter is specified,
+        the elements in the list are sorted based on it's value. The
+        orderByDescending value must be a boolean string. The elements in the
+        list are sorted ascending if false and descending if true.
```
jjaco2: Use the orderBy and orderByDescending fields to sort lists of metadata objects.

The rest of this, I think, could be moved to a parameter description. I've made a comment below showing what this might look like.

file/nifile.yml
```diff
@@ -0,0 +1,816 @@
...
+          description: >-
+            How many files to return in the result, or 0 to use a default
+            defined by the service
+          type: integer
+          x-ni-data-type: U32
+          default: 0
+          maximum: 10000
+        - in: query
+          name: orderBy
+          description: The name of the metadata key to sort by
```
jjaco2: Possible reworking of above description into parameter description here: The name of the metadata key to sort by. This value should be the name of a metadata key. Elements in a list are in ascending order by default.

file/nifile.yml
```diff
@@ -0,0 +1,816 @@
...
+          type: string
+          enum:
+            - created
+            - id
+            - size
+        - in: query
+          name: orderByDescending
+          description: Whether to sort descending instead of ascending
```
jjaco2: Possible reworking of above description into parameter description here:  Specifies whether to sort in descending order instead of ascending.

[I think the rest of the longer description from above goes without saying given the data type + description for this parameter.]

file/nifile.yml
```diff
@@ -0,0 +1,816 @@
...
+          type: boolean
+          default: false
+        - in: query
+          name: id
+          description: Comma-separated list of file IDs to restrict to in the results
```
jjaco2: > to restrict to in the results

Maybe "to search by?"

file/nifile.yml
```diff
@@ -0,0 +1,816 @@
...
+      description: >-
+        Downloads a file from the Skyline File service. The file is returned in
+        a single HTTP response. The file's name metadata property is appended
+        to the response header so that the file is downloaded with the
+        appropriate file name.When the the inline query parameter is not
+        specified or is set to false, the Content-Disposition header is set to
+        'attachment' and the MIME type is set to binary so that the browser
+        handles the response as a file download.  When the inline query
+        parameter is set to true, the Content-Disposition header and MIME type
+        remain unset, and the file contents must be handled by the caller.
```
jjaco2: Returns a file from the SystemLink [I think] File service in an HTTP response.

I think the next sentence (about retaining the name) goes without saying, and the sentences that follow can be moved to parameter descriptions.

file/nifile.yml
```diff
@@ -0,0 +1,816 @@
...
+      operationId: ReceiveFile
+      x-ni-operation: downloadData
+      x-ni-privilege: DownloadFiles
+      x-ni-request-variables:
+        - REQUEST_METHOD
+      parameters:
+        - in: query
+          name: inline
+          description: Whether to return the file data inline rather than as an attachment.
```
jjaco2: When this value is true, the file contents return inline for the caller to handle. When this value is not specified or false, the browser handles the response as a file download.  

Note: I removed parts about the Content-Disposition header and MIME type, because it doesn't seem like these affect end users much. But I could be wrong. Feel free to add it back in.

file/nifile.yml
```diff
@@ -0,0 +1,816 @@
...
+        default:
+          $ref: '#/responses/Error'
+  /v1/service-groups/Default/files/{id}/update-metadata:
+    parameters: 
+      - $ref: '#/parameters/id'
+    post:
+      tags: [files]
+      summary: Update file metadata properties
+      description: >-
+        Updates an existing file's metadata with the supplied metadata
```
jjaco2: supplied --> specified

file/nifile.yml
```diff
@@ -0,0 +1,816 @@
...
+        properties. The ReplaceExisting value determines whether the current
+        list should be entirely replaced by the supplied list, or if entries
+        should be added to (or updated) in the existing set of file properties.
```
jjaco2: Delete details about ReplaceExisting value -- that can be covered below

file/nifile.yml
```diff
@@ -0,0 +1,816 @@
...
+          schema:
+            description: The file's metadata and options for updating it
+            title: UpdateMetadataRequest
+            type: object
+            properties:
+              replaceExisting:
+                description: >-
+                  Whether the current list should be entirely replaced by the
+                  supplied list, or if entries should be added to (or updated)
+                  in the existing set of file properties.
```
jjaco2: Determines whether the current list should be entirely replaced by the specified list.

file/nifile.yml
```diff
@@ -0,0 +1,816 @@
...
+        - in: query
+          name: skip
+          description: How many files to skip in the result when paging.  New in version 2 of this operation.
+          type: integer
+          x-ni-data-type: U32
+          default: 0
+        - in: query
+          name: take
+          description: >-
+            How many files to return in the result, or 0 to use a default.  New in version 2 of this operation.
```
jjaco2, line 742: > or 0 to use a default.

Does that mean if you put 0, it'll default to 1000?

jjaco2 (commented):
Mostly minor things, edits for minimalism, and relocating some descriptions into parameter descriptions. 

A couple additional things to note throughout:

- A lot of my notes are on content that shows up multiple times, but I only comment on the first instance.
- (Super minor, don't have to fix this right now) We typically leave off leading articles "A; The" and periods at the end of things that aren't sentences (like some parameter descriptions)

file/nifile.yml
BKnight760: Yes.  A value of 0 will return 1000 files (the service-defined default).  We've changed this in other services to use -1 instead of 0 to indicate the default since taking 0 results is a valid use-case in some services.

pvallone (approved):
You mentioned that this had already been reviewed, so I'm happy with this so long as Jenna is.

jjaco2 (approved):

PR #9, opened 2018-09-07
Add get routes for results and steps

BKnight760:

This adds two new GET routes to the API to make it more RESTful.  Previously, there were no GET routes at all.  These endpoints call through to the underlying query handler, with no filters applied.  It allows for basic skip/take without filtering.

Updating the public documentation to match the implementation of the backend.

Tested the 2 new routes as well as the base permissions routes.  Verified they behave as expected.


spanglerco (approved):
I don't think we need to add the two new operations and could reuse the query ones. But I'm okay doing that in a separate step to consolidate those.

PR #37, opened 2019-01-11
Tag: Fix codegen when using 2.4.0 swagger-codegen

spanglerco:

After upgrading to [swagger-codegen](https://github.com/swagger-api/swagger-codegen/) [2.4.0](https://github.com/swagger-api/swagger-codegen/releases/tag/v2.4.0), it appears that it no longer generates models that are missing a type and instead require we include `type: object`. Additionally, it's having trouble generating arrays of inlined objects, resulting in things like `List<Object>` instead of it being a `List<MyModel>`. The workaround is to pull out every inline object of an array into a definition. The resulting codegen is identical to what was produced by swagger-codegen 2.3.1.

We're also using this opportunity to rename some of the selection API models to be more consistent, which does result in a codegen change.

This can be seen in https://editor.swagger.io as well, which appears to be using 2.4.0 server-side.

This fixes our ability to generate client code for the tag API using the latest 2.x swagger-codegen.

Generated a C# client using swagger-codegen 2.4.0 with these changes and verified the resulting code is identical to what it was before this change when using swagger-codegen 2.3.1. I also generated a Python client from https://editor.swagger.io and verified the generated code appears correct as well.

### DIFF
--- a/tag/nitag.yml
+++ b/tag/nitag.yml
@@ -43,6 +43,7 @@ responses:
     description: List of tags, including their current and aggregate values
     schema:
       title: Get Tags With Values Response
+      type: object
       properties:
         totalCount:
           description: Total number of tags which match a given query
@@ -56,6 +57,7 @@ responses:
     description: List of tags, excluding current and aggregate values
     schema:
       title: Get Tags Response
+      type: object
       properties:
         totalCount:
           type: integer
@@ -81,6 +83,7 @@ responses:
     description: List of subscription updates
     schema:
       title: V1 Get Subscription Updates Response
+      type: object
       properties:
         subscriptionUpdates:
           description: Subscription updates
@@ -96,6 +99,7 @@ responses:
     description: List of subscription updates
     schema:
       title: V2 Get Subscription Updates Response
+      type: object
       properties:
         subscriptionUpdates:
           description: Subscription updates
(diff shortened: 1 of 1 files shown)


tag/nitag.yml
```diff
@@ -1503,6 +1452,15 @@ definitions:
     example:
       type: DOUBLE
       value: "3.14"
+  TagPathsAndCountValues:
```
pvallone: Ditto - "TagPathWithCountAggregate". In this case we probably want the title to be singular as well. Again, feel free to drop this if there's pressure to just get the codegen fixed quickly.

tag/nitag.yml
```diff
@@ -1525,6 +1483,66 @@ definitions:
...
+  TagPathWithMinAggregate:
+    type: object
+    title: Tag Path with Min Aggregate
+    properties:
+      path:
+        type: string
+      min:
+        description: Minimum value from the tag's aggregates
+        type: string
+  TagPathAndMeanValue:
```
pvallone: Could we name this "TagPathWithMeanAggregate" instead? That would be more consistent with the other new types. I realize that the titles were already inconsistent, and that's what gave rise to the type names. Feel free to drop if there's pressure to just get the codegen fixed quickly.

pvallone (dismissed):

pvallone (approved):

PR #38, opened 2019-01-28
Add alarm swagger doc

pvallone:

This PR adds a Swagger doc for the alarm service.

This PR is in support of F21323: Swagger for Alarms, Tag Rules, and Calibration rules, which is green for SystemLink 19.0.

Hand testing of each route against SLS.

### DIFF
--- a/alarm/nialarm.yml
+++ b/alarm/nialarm.yml
@@ -1,0 +1,687 @@
+swagger: '2.0'
+info:
+  version: '1'
+  title: Alarm Web Service
+  description: SystemLink Alarm Service HTTP API
+  contact:
+    name: National Instruments
+    url: 'https://www.ni.com/systemlink'
+    email: support@ni.com
+basePath: /nialarm
+consumes:
+  - application/json
+produces:
+  - application/json
+securityDefinitions:
+  basicAuth:
+    type: basic
+  cookieAuth:
+    type: apiKey
+    in: header
+    name: Cookie
+security:
+  - basicAuth: []
+  - cookieAuth: []
+x-ni-routing-key: Skyline.AlarmInstance
+definitions:
+  Error:
+    description: Contains error information
+    type: object
+    properties:
+      name:
+        description: String error code
+        type: string
+      code:
+        description: Numeric error code
+        type: integer
+      message:
+        description: Complete error message
+        type: string
+      args:
+        description: Positional argument values for the error code
+        type: array
+        items:
+          type: string
+    example:
+      name: AlarmInstance.InstanceNotFound
+      code: -253006
+      message: An instance with instanceId '5c2cf7e0e0d64403b486fcb4' was not found.
+      args: []
+  Operation:
+    description: Operation provided by the API
+    type: object
+    properties:
+      available:
+        type: boolean
+        description: Whether the operation is available to the caller
+      version:
+        type: integer
+        description: Version of the available operation
+    required: [available]
+    example:
+      available: true
+      version: 1
+  V1Operations:
+    title: V1 Operations
+    description: V1 operations
+    type: object
+    properties:
+      operations:
+        description: >-
+          Available operations in the v1 version of the API:
+          - deleteAlarmInstances: The ability to delete alarm instances
+
+          - readAlarmInstances: The ability to query for alarm instances
+
+          - writeAlarmInstances: The ability to create and modify alarm instances
+
+        type: object
+        properties:
+          deleteAlarmInstances:
+            $ref: '#/definitions/Operation'
+          readAlarmInstances:
+            $ref: '#/definitions/Operation'
+          writeAlarmInstances:
+            $ref: '#/definitions/Operation'
+  AlarmInstance:
+    title: Alarm Instance
+    description: An individual instance of an alarm. The lifecycle of an instance of an alarm begins the first time it is triggered and ends when
+      it has been both acknowledged and cleared. The services enforces the invariant that there can be at most one *active*/true instance of an alarm
+      with a given *alarmId*.
+    type: object
+    required:
+      - alarmId
+    properties:
+      instanceId:
+        description: The unique ID of a particular instance of an alarm.
+        type: string
+        example: 5c33c212e0d6444320d9a9f4
+      alarmId:
+        description: A value meant to uniquely identify the particular process or condition tracked by an alarm. For example, alarm instances created by the Tag Rule Engine Service have their *alarmIds* set to the concatenation of the path of the tag which caused the rule to be triggered and the ID of the rule. CRIO1.System.Health.DiskSpaceLow.Bdd9u!4aMd!$pYrf*CnaIZ2tbu$-Ct%? is an example of such an *alarmId*.
+        type: string
+        example: CRIO1.System.Health.DiskSpaceUsePercentage.Bdd9u!4aMd!$pYrf*CnaIZ2tbu$-Ct%?
+      active:
+        description: Whether or not the alarm instance is active. An active alarm is one that is considered to be still deserving of human or automated attention. Alarm instances always begin life with *active*/true. This field will be automatically set to false when the *clear* and *acknowledged* fields are set to true.
+        type: boolean
+        example: true
+      clear:
+        description: When set to true, the condition that triggered the alarm no longer matches the trigger condition. When an alarm instance is created, *clear* is initially set to false. The creator of the alarm (typically a rule engine or application element of some sort) may determine that the trigger condition no longer applies, and may send an update, clearing the condition. Clearing the alarm does not deactivate the alarm, unless the alarm had previously been acknowledged. As long as the alarm is *active*/true, the alarm may transition from *clear*/true to *clear*/false (or vice versa) multiple times.
+        type: boolean
+        example: false
+      acknowledged:
+        description: When set to true, the alarm has been acknowledged by an alarm handler, which is typically a human. Alarm instances always begin life with *acknowledged*/false. Acknowledging an alarm will not affect the *active* flag unless *clear* is also true. Once both *clear* and *acknowledged* are true, the alarm will become inactive (*active*/false). This field is automatically reset to false when an alarm instance's *highestSeverityLevel* field increases.
+        type: boolean
+        example: false
+      acknowledgedAt:
+        description: The time at which a given alarm instance was acknowledged. This field will be cleared when an alarm instance's *highestSeverityLevel* field increases.
+        type: string
+        example: '2019-01-14T23:15:53.7270539Z'
+      acknowledgedBy:
+        description: Description of who acknowledged a given alarm instance. This field will be cleared when an alarm instance's *highestSeverityLevel* field increases.
+        type: string
+        example: Daffy Duck
+      occurredAt:
+        description: The time at which a given alarm instance was created
+        type: string
+        example: '2019-01-14T23:15:53.7270539Z'
+      updatedAt:
+        description: The time at which a given alarm instance was last updated
+        type: string
+        example: '2019-01-14T23:15:53.7270539Z'
+      createdBy:
+        description: String identifying who or what created a given alarm instance. This is usually used to identify the particular
+          rule engine which requested that a given alarm instance be created.
+        type: string
+        example: TagRuleEngine
+      transitions:
+        description: A collection of transitions for a given alarm instance. Transitions record changes to an alarm instance's *clear* flag, as well as increases or decreases in severity. The service stores the first transition and the most recent one hundred other transitions.
+        type: array
+        items:
+          $ref: '#/definitions/AlarmTransition'
+      transitionOverflowCount:
+        description: The number of transitions which overflowed a given alarm instance's *transition's* field
+        type: integer
+        example: 0
+      currentSeverityLevel:
+        description: The current severity level of a given alarm instance
+        type: integer
+        example: 2
+      highestSeverityLevel:
+        description: The highest severity level that a given alarm instance has ever been in
+        type: integer
+        example: 4
+      channel:
+        description: An identifier for the tag or resource associated with a given alarm instance
+        type: string
+        example: CRIO1.System.Health.DiskSpaceUsePercentage
+      condition:
+        description: A description of the condition associated with a given alarm instance
+        type: string
+        example: Greater than 90
+      displayName:
+        description: Optional display name for a given alarm instance. Display names do not need to be unique.
+        type: string
+        example: Low disk space on tester-1.
+      description:
+        description: Optional description text for a given alarm instance
+        type: string
+        example: Disk space on tester-1 is high. Disk usage was 95% when this alarm was created.
+      keywords:
+        description: Keywords associated with an alarm instance. Alarm instances can be tagged with keywords to make it easier to find them with queries.
+        type: array
+        items:
+          type: string
+        example:
+          - keyword1
+          - keyword2
+      notes:
+        description: A collection of notes for a given alarm instance. Notes are set by humans to record information about an alarm after it has been investigated.
+        type: array
+        items:
+          $ref: '#/definitions/AlarmNote'
+      properties:
+        description: Key-value-pair metadata associated with an alarm instance. Alarm instances can be tagged with properties so that they can be more easily found with queries. Additionally, alarm instance properties are used to expand message templates in the notification strategies associated with a given alarm instance.
+        type: object
+        additionalProperties:
+          type: string
+        example:
+          key1: value1
+      resourceType:
+        description: The type of resource associated with a given alarm instance
+        type: string
+        example: Tag
+  AlarmNote:
+    title: Alarm Note
+    description: Information about a particular alarm instance, such as a description of an investigation into an alarm's root cause.
+    required:
+      - note
+    properties:
+      note:
+        description: Information about a particular alarm instance, such as a description of an investigation into an alarm's root cause.
+        type: string
+        example: Temperature sensor was faulty
+      createdAt:
+        description: The time at which the note was created
+        type: string
+        example: '2019-01-14T23:15:53.7270539Z'
+      user:
+        description: The user who created the note
+        type: string
+        readOnly: true
+        example: Philip
+  AlarmTransitionType:
+    title: Alarm Transition Type
+    type: string
+    enum: [SET, CLEAR]
+  AlarmTransition:
+    title: Alarm Transition
+    description: A transition within a given alarm instance. Transitions can record changes to an alarm's *clear* field, as well as an alarm increasing or decreasing in severity.
+    type: object
+    required:
+      - transitionType
+    properties:
+      transitionType:
+        $ref: '#/definitions/AlarmTransitionType'
+      occurredAt:
+        description: The time at which the transition occurred.
+        type: string
+        example: '2019-01-14T23:15:53.7270539Z'
+      severityLevel:
+        description: The severity of the transition. Valid values for CLEAR transitions are [-1,-1]. Valid values for SET transitions are [1, 4].
+        type: integer
+        example: 2
+        default: 2
+        minimum: -1
+        maximum: 4
+      notificationStrategyIds:
+        description: The IDs of the notification strategies which should be triggered if this request results in an alarm instance being created or transitioning to a new highest severity
+        type: array
+        items:
+          type: string
+        example: []
+      condition:
+        description: The condition associated with the transition
+        type: string
+        example: Greater than 90
+      shortText:
+        description: Optional additional description of the condition
+        type: string
+        example: Low disk space
+      detailText:
+        description: Optional additional description of the condition
+        type: string
+        example: Disk usage on CRIO1 is 95% (lower than configured threshold of 90%)
+      keywords:
+        description: Keywords associated with a transition
+        type: array
+        items:
+          type: string
+        example:
+          - keyword1
+          - keyword2
+      properties:
+        description: Key-value-pair metadata associated with a transition
+        type: object
+        additionalProperties:
+          type: string
+        example:
+          key1: value1
+  SubQuery:
+    title: Nested Query Body
+    type: object
+    properties:
+      alarmId:
+        type: string
+        example: System.Health.DiskSpaceLow.Bdd9u!4aMd!$pYrf*CnaIZ2tbu$-Ct%?
+      displayName:
+        type: string
+        description: Display name query. The service will return alarm instances whose *displayName* fields contain the display name query string, without taking case into account.
+        example: Low disk space on CRIO1
+      active:
+        type: boolean
+        example: true
+      clear:
+        type: boolean
+        example: false
+      acknowledged:
+        type: boolean
+        example: false
+      acknowledgedAtMin:
+        type: string
+        description: Acknowledged at minimum date query. The service will return instances which were acknowledged between *acknowledgedAtMin* and *acknowledgedAtMax*, inclusive.
+        example: '2019-01-14T23:15:53.7270539Z'
+      acknowledgedAtMax:
+        type: string
+        description: Acknowledged at maximum date query. The service will return instances which were acknowledged between *acknowledgedAtMin* and *acknowledgedAtMax*, inclusive.
+        example: '2019-01-15T23:15:53.7270539Z'
+      occurredAtMin:
+        type: string
+        description: Occurred at minimum date query. The service will return instances which occurred between *occurredAtMin* and *occurredAtMax*, inclusive.
+        example: '2019-01-14T23:15:53.7270539Z'
+      occurredAtMax:
+        type: string
+        description: Occurred at maximum date query. The service will return instances which occurred between *occurredAtMin* and *occurredAtMax*, inclusive.
+        example: '2019-01-15T23:15:53.7270539Z'
+      currentSeverityLevelMin:
+        type: integer
+        description: Current severity minimum query. The service will return instances whose current severity is between *currentSeverityLevelMin* and *currentSeverityLevelMax*, inclusive.
+        example: 1
+      currentSeverityLevelMax:
+        type: integer
+        description: Current severity maximum query. The service will return instances whose current severity is between *currentSeverityLevelMin* and *currentSeverityLevelMax*, inclusive.
+        example: 4
+      highestSeverityLevelMin:
+        type: integer
+        description: Highest severity minimum query. The service will return instances whose highest severity is between *highestSeverityLevelMin* and *highestSeverityLevelMax*, inclusive.
+        example: 1
+      highestSeverityLevelmax:
+        type: integer
+        description: Highest severity maximum query. The service will return instances whose highest severity is between *highestSeverityLevelMin* and *highestSeverityLevelMax*, inclusive.
+        example: 4
+      createdBy:
+        type: string
+        description: Created by query. The service will return instances whose *createdBy* fields are equal to the query, taking case into account.
+        example: TagRuleEngine
+      channel:
+        type: string
+        description: Channel query. The service will return instances whose *channel* fields are equal to the query, taking case into account. Supports wildcard match checking uses glob-style wildcards.
+        example: '*.System.Health.DiskSpaceUsePercentage'
+      properties:
+        type: object
+        description: Property query. The service will return instances whose *properties* fields contain all of the specified key/value pairs.
+        example:
+          key1: value1
+      keywords:
+        type: object
+        description: Keyword query. The service will return instances whose *keywords* fields contain all of the specified keywords.
+        example:
+          - keyword1
+          - keyword2
+  QueryRequest:
+    title: Query Request
+    allOf:
+    - $ref: '#/definitions/SubQuery'
+    - type: object
+      properties:
+        subQueries:
+          description: Additional queries. The service will return instances which match any of the queries in the subQueries array. If a top-level query was provided, the top-level query will be AND'ed with each of the sub queries.
+          type: array
+          items:
+            $ref: '#/definitions/SubQuery'
+        returnMostRecentlyOccurredOnly:
+          description: For the instances matched by this query, only return the most recent instance for each *alarmId*. In this context, recency is based on when the alarm instance occurred, not when it was last updated.
+          type: boolean
+        orderBy:
+          description: Whether or not results should be sorted in ascending or descending order relative to the alarm instance's *updatedAt* field.
+          type: string
+          enum: [DATE_UPDATED_FORWARD, DATE_UPDATED_BACKWARD]
+        skip:
+          description: Number of entries to skip in the response list. Typically used in combination with the take parameter to implement pagination.
+          type: integer
+        take:
+          description: Number of entries to include in the response list. Typically used in combination with the skip parameter to implement pagination.
+          type: integer
+
+responses:
+  Error:
+    description: Error
+    schema:
+      description: Error response
+      title: ErrorResponse
+      type: object
+      properties:
+        error:
+          $ref: '#/definitions/Error'
+  QueryResponse:
+    description: Query response
+    schema:
+      description: Query response
+      title: QueryResponse
+      type: object
+      properties:
+        totalCount:
+          description: The total number of alarm instances which matched your query
+          type: integer
+          example: 1
+        filterMatches:
+          description: Array containing the alarm instances which matched your query
+          type: array
+          items:
+            $ref: '#/definitions/AlarmInstance'
+  Unauthorized:
+    description: Not authorized
+    headers:
+      WWW_Authenticate:
+        description: Information for how to authenticate
+        type: string
+paths:
+  /:
+    get:
+      tags: [versioning]
+      summary: API information
+      description: Returns information about API versions and available operations.
+      operationId: RootEndpoint
+      x-ni-request-all-privileges: true
+      # Explicitly mark security as an empty array - this route does not require any privileges.
+      # Marking it this way prevents it from inheriting the top-level security settings.
+      security: []
+      responses:
+        200:
+          description: OK
+          schema:
+            description: Version information
+            title: RootEndpointResponse
+            type: object
+            properties:
+              v1:
+                $ref: '#/definitions/V1Operations'
+              version:
+                description: Implementation version of the web service
+                type: string
+  /{version}:
+    parameters:
+      - in: path
+        name: version
+        description: Version of the API to retrieve operations.
+        type: string
+        required: true
+    get:
+      tags: [versioning]
+      summary: API version information
+      description: Returns available operations for a single version of the API.
+      operationId: RootEndpointWithVersion
+      x-ni-request-all-privileges: true
+      # Explicitly mark security as an empty array - this route does not require any privileges.
+      # Marking it this way prevents it from inheriting the top-level security settings.
+      security: []
+      responses:
+        200:
+          description: OK
+          schema:
+            $ref: '#/definitions/V1Operations'
+        404:
+          description: Not Found
+          schema:
+            $ref: '#/definitions/Error'
+  /v1/acknowledge-instances:
+    post:
+      tags: [alarmInstances]
+      summary: Acknowledge alarm instances
+      description: Acknowledges one or more alarm instances, optionally forcing them clear and adding notes to them.
+      operationId: AcknowledgeInstances
+      parameters:
+        - in: body
+          name: Request body
+          schema:
+            type: object
+            title: Request Body
+            required:
+              - alarmIds
+            properties:
+              alarmIds:
+                description: The *alarmIds* of the alarm instances which should be acknowledged.
+                type: array
+                items:
+                  type: string
+                example: [5c40ec55e0d6441168b4c543, 5c40ec55e0d6441168b4c543]
+              forceClear:
+                description: Whether or not the affected alarm instances should have their *clear* field set to true.
+                type: boolean
+                example: false
+                default: false
+              notes:
+                description: Notes which should be added to the alarm instances.
+                type: array
+                items:
+                  $ref: '#/definitions/AlarmNote'
+      responses:
+        200:
+          description: OK
+          schema:
+            type: object
+            properties:
+              acknowledged:
+                description: The *alarmIds* which were successfully acknowledged.
+                type: array
+                items:
+                  type: string
+                example: [5c40ec55e0d6441168b4c543, 5c40ec55e0d6441168b4c543]
+              alreadyAcknowledged:
+                description: The *alarmIds* which had already been previously acknowledged.
+                type: array
+                items:
+                  type: string
+                example: [5c40ec55e0d6441168b4c541]
+              notFound:
+                description: A collection of *alarmIds* for which no active alarm instances were found.
+                type: array
+                items:
+                  type: string
+                example: [5c40ec55e0d6441168b4c549]
+        401:
+          description: Unauthorized
+          schema:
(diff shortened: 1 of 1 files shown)


alarm/nialarm.yml
```diff
@@ -0,0 +1,686 @@
...
+      version: 1
+  V1Operations:
+    title: V1 Operations
+    description: V1 operations
+    type: object
+    properties:
+      operations:
+        description: >-
+          Available operations in the v1 version of the API:
+          - deleteAlarmInstances: The ability to delete alarm instances
```
spanglerco, line 68: The extra blank line needs to go before the bullet rather than after in order to get a newline in the output.

alarm/nialarm.yml
```diff
@@ -0,0 +1,686 @@
...
+    type: object
+    required:
+      - alarmId
+    properties:
+      instanceId:
+        description: The unique ID of a particular instance of an alarm.
+        type: string
+        example: 5c33c212e0d6444320d9a9f4
+      alarmId:
+        description: A value meant to uniquely identify the particular process or condition tracked by an alarm. For example, alarm instances created by the Tag Rule Engine Service have their *alarmIds* set to the concatenation of the path of the tag which caused the rule to be triggered and the ID of the rule. CRIO1.System.Health.DiskSpaceLow.Bdd9u!4aMd!$pYrf*CnaIZ2tbu$-Ct%? is an example of such an *alarmId*.
```
spanglerco: Feel free to wrap these long descriptions onto multiple lines :)

>  CRIO1.System.Health.DiskSpaceLow.Bdd9u!4aMd!$pYrf*CnaIZ2tbu$-Ct%? is an example of such an *alarmId*.

This seems redundant given the example attribute, unless you are wanting to be explicit about how the example is for tag rule engine?

alarm/nialarm.yml
```diff
@@ -0,0 +1,686 @@
...
+        type: string
+        example: CRIO1.System.Health.DiskSpaceUsePercentage.Bdd9u!4aMd!$pYrf*CnaIZ2tbu$-Ct%?
+      active:
+        description: Whether or not the alarm instance is active. An active alarm is one that is considered to be still deserving of human or automated attention. Alarm instances always begin life with *active*/true. This field will be automatically set to false when the *clear* and *acknowledged* fields are set to true.
```
spanglerco: > active/true

Is this a standard format for referring to the states, or should it be active=true? Same goes for other descriptions like in clear below.

alarm/nialarm.yml
```diff
@@ -0,0 +1,686 @@
...
+        type: boolean
+        example: false
+      acknowledged:
+        description: When set to true, the alarm has been acknowledged by an alarm handler, which is typically a human. Alarm instances always begin life with *acknowledged*/false. Acknowledging an alarm will not affect the *active* flag unless *clear* is also true. Once both *clear* and *acknowledged* are true, the alarm will become inactive (*active*/false). This field is automatically reset to false when an alarm instance's *highestSeverityLevel* field increases.
+        type: boolean
+        example: false
+      acknowledgedAt:
+        description: The time at which a given alarm instance was acknowledged. This field will be cleared when an alarm instance's *highestSeverityLevel* field increases.
```
spanglerco: For date and time properties, add a `format: iso-date-time` attribute. I think we've also started explicitly stating in the description that it's in ISO-8601 format.

alarm/nialarm.yml
```diff
@@ -0,0 +1,686 @@
...
+          type: string
+        example:
+          key1: value1
+      resourceType:
+        description: The type of resource associated with a given alarm instance
+        type: string
+        example: Tag
+  AlarmNote:
+    title: Alarm Note
+    description: Information about a particular alarm instance, such as a description of an investigation into an alarm's root cause.
```
spanglerco: Do you think this description should have something a little more to distinguish it from the description of the note property inside it? Maybe something as simple as `Contains information ...`?

alarm/nialarm.yml
```diff
@@ -0,0 +1,686 @@
...
+        type: string
+        example: Temperature sensor was faulty
+      createdAt:
+        description: The time at which the note was created
+        type: string
+        example: '2019-01-14T23:15:53.7270539Z'
+      user:
+        description: The user who created the note
+        type: string
+        readOnly: true
```
spanglerco, line 233: Neat, I didn't know about readOnly. I wonder if we should be using that in more places than we are across all our APIs.

alarm/nialarm.yml
```diff
@@ -0,0 +1,686 @@
...
+        example: Philip
+  AlarmTransitionType:
+    title: Alarm Transition Type
+    type: string
+    enum: [SET, CLEAR]
+  AlarmTransition:
+    title: Alarm Transition
+    description: A transition within a given alarm instance. Transitions can record changes to an alarm's *clear* field, as well as an alarm increasing or decreasing in severity.
```
spanglerco: You can probably leave out the "can" in the sentence, because transitions do record changes.

alarm/nialarm.yml
```diff
@@ -0,0 +1,686 @@
...
+        description: The time at which the transition occurred.
+        type: string
+        example: '2019-01-14T23:15:53.7270539Z'
+      severityLevel:
+        description: The severity of the transition. Valid values for CLEAR transitions are [-1,-1]. Valid values for SET transitions are [1, 4].
+        type: integer
+        example: 2
+        default: 2
+        minimum: -1
+        maximum: 4
```
spanglerco: I thought we didn't enforce a maximum at the alarm web service layer.

alarm/nialarm.yml
```diff
@@ -0,0 +1,686 @@
...
+      notificationStrategyIds:
+        description: The IDs of the notification strategies which should be triggered if this request results in an alarm instance being created or transitioning to a new highest severity
+        type: array
+        items:
+          type: string
+        example: []
```
spanglerco, line 268: Should we remove this example if it's empty?

alarm/nialarm.yml
```diff
@@ -0,0 +1,686 @@
...
+      condition:
+        description: The condition associated with the transition
```
spanglerco: Should this description more closely match AlarmInstance.condition saying this field is a description of the condition rather than the condition itself?

alarm/nialarm.yml
```diff
@@ -0,0 +1,686 @@
...
+        type: string
+        example: Greater than 90
+      shortText:
+        description: Optional additional description of the condition
+        type: string
+        example: Low disk space
+      detailText:
+        description: Optional additional description of the condition
```
spanglerco: I think it would be useful to have descriptions differentiate between shortText and detailText and how they're used.

alarm/nialarm.yml
```diff
@@ -0,0 +1,686 @@
...
+      keywords:
+        description: Keywords associated with a transition
+        type: array
+        items:
+          type: string
+        example:
+          - keyword1
+          - keyword2
+      properties:
+        description: Key-value-pair metadata associated with a transition
```
spanglerco: Similarly, including what a transition's keywords and properties are used for would be helpful to a user of the API.

alarm/nialarm.yml
```diff
@@ -0,0 +1,686 @@
...
+        type: object
+        additionalProperties:
+          type: string
+        example:
+          key1: value1
+  SubQuery:
+    title: Nested Query Body
```
spanglerco: It can make it a little confusing to have the title too different from the model name. For example, the codegen will ignore the title and generate a model named SubQuery.

Also, should the model have a description? Several of its properties are missing descriptions as well.

alarm/nialarm.yml
```diff
@@ -0,0 +1,686 @@
...
+        type: string
+        example: 5c33c212e0d6444320d9a9f4
+      alarmId:
+        description: A value meant to uniquely identify the particular process or condition tracked by an alarm. For example, alarm instances created by the Tag Rule Engine Service have their *alarmIds* set to the concatenation of the path of the tag which caused the rule to be triggered and the ID of the rule. CRIO1.System.Health.DiskSpaceLow.Bdd9u!4aMd!$pYrf*CnaIZ2tbu$-Ct%? is an example of such an *alarmId*.
+        type: string
+        example: CRIO1.System.Health.DiskSpaceUsePercentage.Bdd9u!4aMd!$pYrf*CnaIZ2tbu$-Ct%?
+      active:
+        description: Whether or not the alarm instance is active. An active alarm is one that is considered to be still deserving of human or automated attention. Alarm instances always begin life with *active*/true. This field will be automatically set to false when the *clear* and *acknowledged* fields are set to true.
+        type: boolean
+        example: true
```
spanglerco: We probably don't need example values for boolean fields, since there are only two choices, unless there's a specific reason we want to call out a set of example values for an object.

alarm/nialarm.yml
```diff
@@ -0,0 +1,686 @@
...
+        example: 1
+      currentSeverityLevelMax:
+        type: integer
+        description: Current severity maximum query. The service will return instances whose current severity is between *currentSeverityLevelMin* and *currentSeverityLevelMax*, inclusive.
+        example: 4
+      highestSeverityLevelMin:
+        type: integer
+        description: Highest severity minimum query. The service will return instances whose highest severity is between *highestSeverityLevelMin* and *highestSeverityLevelMax*, inclusive.
+        example: 1
+      highestSeverityLevelmax:
```
spanglerco: highestSeverityLevelmax -> highestSeverityLevel**M**ax

alarm/nialarm.yml
```diff
@@ -0,0 +1,686 @@
...
+      createdBy:
+        type: string
+        description: Created by query. The service will return instances whose *createdBy* fields are equal to the query, taking case into account.
+        example: TagRuleEngine
+      channel:
+        type: string
+        description: Channel query. The service will return instances whose *channel* fields are equal to the query, taking case into account. Supports wildcard match checking uses glob-style wildcards.
+        example: '*.System.Health.DiskSpaceUsePercentage'
+      properties:
+        type: object
```
spanglerco, line 382: Should this have an additionalProperties attribute?

alarm/nialarm.yml
```diff
@@ -0,0 +1,686 @@
...
+        description: Property query. The service will return instances whose *properties* fields contain all of the specified key/value pairs.
+        example:
+          key1: value1
+      keywords:
+        type: object
```
spanglerco: Should this be an array of strings?

alarm/nialarm.yml
```diff
@@ -0,0 +1,686 @@
...
+        description: Keyword query. The service will return instances whose *keywords* fields contain all of the specified keywords.
+        example:
+          - keyword1
+          - keyword2
+  QueryRequest:
+    title: Query Request
```
spanglerco, line 399: Ditto on this model not having a description.

alarm/nialarm.yml
```diff
@@ -0,0 +1,686 @@
...
+          description: For the instances matched by this query, only return the most recent instance for each *alarmId*. In this context, recency is based on when the alarm instance occurred, not when it was last updated.
+          type: boolean
+        orderBy:
+          description: Whether or not results should be sorted in ascending or descending order relative to the alarm instance's *updatedAt* field.
+          type: string
+          enum: [DATE_UPDATED_FORWARD, DATE_UPDATED_BACKWARD]
+        skip:
+          description: Number of entries to skip in the response list. Typically used in combination with the take parameter to implement pagination.
+          type: integer
+          example: 0
```
spanglerco: Probably don't need an example if it matches the default value. Applies to `take` as well.

alarm/nialarm.yml
```diff
@@ -0,0 +1,686 @@
...
+    description: Error
+    schema:
+      description: Error response
+      title: ErrorResponse
+      type: object
+      properties:
+        error:
+          $ref: '#/definitions/Error'
+  QueryResponse:
+    description: Query response
```
spanglerco: This description is meant to describe when the response is returned, and would typically be something like `OK` or `Success`.

alarm/nialarm.yml
```diff
@@ -0,0 +1,686 @@
...
+          description: OK
+          schema:
+            $ref: '#/definitions/V1Operations'
+        404:
+          description: Not Found
+          schema:
+            $ref: '#/definitions/Error'
+  /v1/acknowledge-instances:
+    post:
+      tags: [alarmInstances]
```
spanglerco: Tags can contain spaces and should be all lowercase, so `alarm instances` would be fine.

alarm/nialarm.yml
```diff
@@ -0,0 +1,686 @@
...
+      summary: Acknowledge alarm instances
+      description: Acknowledges one or more alarm instances, optionally forcing them clear and adding notes to them.
+      operationId: AcknowledgeInstances
+      parameters:
+        - in: body
+          name: Request body
+          schema:
+            type: object
+            title: Request Body
```
spanglerco: This should be a more specific title for use in codegen.

alarm/nialarm.yml
```diff
@@ -0,0 +1,686 @@
...
+                default: false
+              notes:
+                description: Notes which should be added to the alarm instances.
+                type: array
+                items:
+                  $ref: '#/definitions/AlarmNote'
+      responses:
+        200:
+          description: OK
+          schema:
```
spanglerco, line 552: This response schema needs a title for codegen.

alarm/nialarm.yml
```diff
@@ -0,0 +1,686 @@
...
+        404:
+          description: Not Found
+          schema:
+            $ref: '#/definitions/Error'
+  /v1/acknowledge-instances:
+    post:
+      tags: [alarmInstances]
+      summary: Acknowledge alarm instances
+      description: Acknowledges one or more alarm instances, optionally forcing them clear and adding notes to them.
+      operationId: AcknowledgeInstances
```
spanglerco, line 521: We've conventionally included x-ni-operation attributes to associate each route with one of the V1Operations in addition to the operationId.

alarm/nialarm.yml
```diff
@@ -0,0 +1,686 @@
...
+      summary: Create or update an alarm instance
+      description: Creates or updates an instance of an alarm based on the requested transition and the state of the current active alarm instance with the given *alarmId*.
+      operationId: CreateOrUpdateInstance
+      parameters:
+        - in: body
+          name: Request body
+          description: Request body. For requests which do not result in a new alarm instance being created, all fields other than *alarmId* and *transition* are ignored.
+          schema:
+            type: object
+            title: Request body
```
spanglerco: Ditto on a more specific title.

alarm/nialarm.yml
```diff
@@ -0,0 +1,686 @@
...
+                description: Keywords associated with an alarm instance. Alarm instances can be tagged with keywords to make it easier to find them with queries.
+                type: array
+                items:
+                  type: string
+                example:
+                  - keyword1
+                  - keyword2
+              properties:
+                description: Key-value-pair metadata associated with an alarm instance. Alarm instances can be tagged with properties so that they can be more easily found with queries. Additionally, alarm instance properties are used to expand message templates in the notification strategies associated with a given alarm instance.
+                type: object
```
spanglerco, line 643: Should this have an additionalProperties attribute?

alarm/nialarm.yml
```diff
@@ -0,0 +1,686 @@
...
+                example:
+                  key1: value1
+          required: true
+      responses:
+        200:
+          description: OK
+          schema:
+            type: object
+            title: ID of the created or modified instance
```
spanglerco: This seems like kind of an oddly specific title. Maybe it could be more symmetric with the request body, like `Create or update instance response`?

alarm/nialarm.yml
```diff
@@ -0,0 +1,686 @@
...
+    parameters:
+      - in: path
+        name: id
+        description: Instance ID
+        type: string
+        required: true
+    get:
+      tags: [alarmInstances]
+      summary: Get an alarm instance
+      description: Gets an alarm instance by its *instanceId*
```
spanglerco: This should be a complete sentence that ends with punctuation.

alarm/nialarm.yml
```diff
@@ -0,0 +1,686 @@
...
+      responses:
+        200:
+          $ref: '#/definitions/AlarmInstance'
```
spanglerco: This is a ref to model, not a response, so the UI shows it correctly, but codegen results in a void method. It needs to be wrapped in a description + schema.

alarm/nialarm.yml
```diff
@@ -0,0 +1,686 @@
...
+          $ref: '#/responses/Error'
+  /v1/instances/{id}:
+    parameters:
+      - in: path
+        name: id
+        description: Instance ID
+        type: string
+        required: true
+    get:
+      tags: [alarmInstances]
```
spanglerco: This route is missing an operationId, so codegen uses the automatic one `V1InstancesIdGet`.

alarm/nialarm.yml
```diff
@@ -0,0 +1,686 @@
...
+      summary: Get an alarm instance
+      description: Gets an alarm instance by its *instanceId*
+      responses:
+        200:
+          $ref: '#/definitions/AlarmInstance'
+    delete:
+      tags: [alarmInstances]
+      summary: Delete an alarm instance
+      description: Deletes an alarm instance by its *instanceId*
```
spanglerco: Ditto on this route not having an operationId + description needs punctuation.

alarm/nialarm.yml
```diff
@@ -0,0 +1,686 @@
...
+            $ref: '#/definitions/Error'
+  /v1/acknowledge-instances:
+    post:
+      tags: [alarmInstances]
+      summary: Acknowledge alarm instances
+      description: Acknowledges one or more alarm instances, optionally forcing them clear and adding notes to them.
+      operationId: AcknowledgeInstances
+      parameters:
+        - in: body
+          name: Request body
```
spanglerco, line 525: Should this parameter be marked as required?

alarm/nialarm.yml
```diff
@@ -0,0 +1,686 @@
...
+      tags: [alarmInstances]
+      summary: Add notes to an alarm instance
+      description: Adds one or more notes to an alarm instance.
+      operationId: AddNotesToInstance
+      parameters:
+        - in: body
+          name: Request body
+          required: true
+          schema:
+            type: object
```
spanglerco, line 722: This object needs a title for the codegen.

alarm/nialarm.yml
```diff
@@ -0,0 +1,686 @@
...
+            required:
+              - notes
+            properties:
+              notes:
+                type: array
```
spanglerco, line 728: It's pretty self-explanatory, but should this property have a description anyway?

alarm/nialarm.yml
```diff
@@ -0,0 +1,686 @@
...
+          description: No Content
+        401:
+          $ref: '#/responses/Unauthorized'
+        default:
+          $ref: '#/responses/Error'
+  /v1/query-instances:
+    post:
+      tags: [alarmInstances]
+      summary: Query alarm instances
+      description: Queries for alarm instances
```
spanglerco: This should be a complete sentence and should probably include an explanation of what happens if the request body is empty.

alarm/nialarm.yml
```diff
@@ -0,0 +1,686 @@
...
+              instanceIds:
+                description: Instance IDs to delete
+                type: array
+                items:
+                  type: string
+                example:
+                  - 5c33c212e0d6444320d9a9f4
+                  - 5c33c212e0d6444320d9a9f5
+      responses:
+        401:
```
spanglerco, line 786: This route is missing any success responses.

spanglerco (changes requested):

alarm/nialarm.yml
pvallone: Ah. I copied this from test monitor, so they probably have the same issue.

alarm/nialarm.yml
pvallone: Wrapped all strings to 110 lines. Got rid of the redundant example.

alarm/nialarm.yml
pvallone: It's the format used in the alarm spec and in documentation Michael had written for the original prototype of the service.

alarm/nialarm.yml
pvallone: Added the attribute to all date and time properties

alarm/nialarm.yml
pvallone: Prefaced it with Contains, but kept the rest. Let me know if you had something else in mind.

alarm/nialarm.yml
pvallone: We don't, but the we only have display strings like "Critical (4)" for these values in the UI, so I wanted to sort of encourage folks to only use those values without actually enforcing a hard limit. Let me know if you think the document should reflect what the API actually allows instead.

alarm/nialarm.yml
pvallone: I'm doing this so that the example request is a valid request which the service accepts. If I don't provide an example, Swagger sticks an array with "string" in it. That isn't a valid BSON ID, so the service returns an error if you send the example request. The request works if I pass in an empty array.

alarm/nialarm.yml
pvallone: Yep, made it so.

alarm/nialarm.yml
pvallone: I added some more detail. Transition properties and keywords aren't really used for anything in the API, they're just optional metadata which folks can add to transitions.

alarm/nialarm.yml
pvallone: I fleshed out the description text a bit more. They aren't really used for anything, they're just optional metadata which folks can add to transitions.

alarm/nialarm.yml
pvallone: Changed the title and added missing descriptions.

alarm/nialarm.yml
pvallone: Removed the examples for boolean fields.

alarm/nialarm.yml
pvallone: Woops! I confess to not having tested all of the query properties.

alarm/nialarm.yml
pvallone: Yep!

alarm/nialarm.yml
pvallone: Added missing additionalProperties attributes

alarm/nialarm.yml
pvallone: Nice, I much prefer them that way. For some reason I thought past guidance had been to camelCase them.

alarm/nialarm.yml
pvallone: Fixed the title

alarm/nialarm.yml
pvallone: I added a period. I don't think I need to do anything else to make it a complete sentence, as I guess the subject is implied. The corresponding strings in the test monitor document are in the same style, anyway.

alarm/nialarm.yml
pvallone: Yep!

alarm/nialarm.yml
pvallone: Added one

alarm/nialarm.yml
pvallone: Added a 204

alarm/nialarm.yml
pvallone: Added the attributes. Do I also need x-ni-privilege attributes, like the test monitor document has?

alarm/nialarm.yml
spanglerco: A strict Swagger codegen could in theory throw an exception trying to use a value greater than 4 since that's what it says is the max. I don't know of any clients specifically that do that, but I believe the Python server codegen does. Since it's more of a "soft" maximum, maybe putting that in the description instead would be sufficient?

alarm/nialarm.yml
spanglerco: That's more of an implementation detail (and gets stripped during build of the version included in /niapis), so I wouldn't worry about it.

alarm/nialarm.yml
pvallone: Removed the maximum and documented the soft maximum in the description.

spanglerco (dismissed):

alarm/nialarm.yml
```diff
@@ -0,0 +1,790 @@
...
+        properties:
+          deleteAlarmInstances:
+            $ref: '#/definitions/Operation'
+          readAlarmInstances:
+            $ref: '#/definitions/Operation'
+          writeAlarmInstances:
+            $ref: '#/definitions/Operation'
+  AlarmInstance:
+    title: Alarm Instance
+    description: An individual instance of an alarm. The lifecycle of an instance of an alarm begins the first
```
caleyburton: For the second sentence, could we rephrase it to the following: "The lifecycle of an alarm" or "The lifecycle of an alarm instance" (whichever is most technically correct)?

alarm/nialarm.yml
```diff
@@ -0,0 +1,790 @@
...
+      time it is triggered and ends when it has been both acknowledged and cleared. The services enforces the
```
caleyburton: "The services enforce" or "The service enforces"?

alarm/nialarm.yml
```diff
@@ -0,0 +1,790 @@
...
+      - alarmId
+    properties:
+      instanceId:
+        description: The unique ID of a particular instance of an alarm.
+        type: string
+        example: 5c33c212e0d6444320d9a9f4
+      alarmId:
+        description: A value meant to uniquely identify the particular process or condition tracked by an
+          alarm. For example, alarm instances created by the Tag Rule Engine Service have their *alarmIds* set
+          to the concatenation of the path of the tag which caused the rule to be triggered and the ID of the
```
caleyburton, line 98: Would "tag path" work here instead of "path of the tag"?

alarm/nialarm.yml
```diff
@@ -0,0 +1,790 @@
...
+          rule.
+        type: string
+        example: CRIO1.System.Health.DiskSpaceUsePercentage.Bdd9u!4aMd!$pYrf*CnaIZ2tbu$-Ct%?
+      active:
+        description: Whether or not the alarm instance is active. An active alarm is one that is considered to
+          be still deserving of human or automated attention. Alarm instances always begin life with
```
caleyburton: "An active alarm deserves human or automated attention."

alarm/nialarm.yml
```diff
@@ -0,0 +1,790 @@
...
+          alarm (typically a rule engine or application element of some sort) may determine that the trigger
+          condition no longer applies, and may send an update, clearing the condition. Clearing the alarm
+          does not deactivate the alarm, unless the alarm had previously been acknowledged. As long as the
+          alarm is *active*/true, the alarm may transition from *clear*/true to *clear*/false (or vice versa)
+          multiple times.
+        type: boolean
+      acknowledged:
+        description: When set to true, the alarm has been acknowledged by an alarm handler, which is typically
+          a human. Alarm instances always begin life with *acknowledged*/false. Acknowledging an alarm will
+          not affect the *active* flag unless *clear* is also true. Once both *clear* and *acknowledged* are
```
caleyburton: Instead of "Once" can we say "When"?

alarm/nialarm.yml
```diff
@@ -0,0 +1,790 @@
...
+          when an alarm instance's *highestSeverityLevel* field increases.
+        type: boolean
+      acknowledgedAt:
+        description: The time at which a given alarm instance was acknowledged. This field will be cleared
+          when an alarm instance's *highestSeverityLevel* field increases.
+        type: string
+        format: iso-date-time
+        example: '2019-01-14T23:15:53.7270539Z'
+      acknowledgedBy:
+        description: Description of who acknowledged a given alarm instance. This field will be cleared when
```
caleyburton: Could we rephrase it to, "An identifier for who acknowledged a given alarm instance."?

alarm/nialarm.yml
```diff
@@ -0,0 +1,790 @@
...
+           an alarm instance's *highestSeverityLevel* field increases.
+        type: string
+        example: Daffy Duck
```
caleyburton, line 133: I love this example. 👍

alarm/nialarm.yml
```diff
@@ -0,0 +1,790 @@
...
+        type: boolean
+      acknowledged:
+        description: When set to true, the alarm has been acknowledged by an alarm handler, which is typically
+          a human. Alarm instances always begin life with *acknowledged*/false. Acknowledging an alarm will
+          not affect the *active* flag unless *clear* is also true. Once both *clear* and *acknowledged* are
+          true, the alarm will become inactive (*active*/false). This field is automatically reset to false
+          when an alarm instance's *highestSeverityLevel* field increases.
+        type: boolean
+      acknowledgedAt:
+        description: The time at which a given alarm instance was acknowledged. This field will be cleared
```
caleyburton: Looking at the other API docs' time-related descriptions, they use something along the lines of "ISO-8601 formatted timestamp specifying X". So if we were to mimic that, we could update the description to "ISO-8601 formatted timestamp specifying when the alarm handler acknowledged the alarm instance." or "ISO-8601 formatted timestamp specifying when the alarm instance was acknowledged." 

This comment goes for the following time-related descriptions (AlarmInstance's occurredAt, updatedAt, createdAt; AlarmTransition's occurredAt; AlarmNote's createdAt) if it makes sense and you agree.

alarm/nialarm.yml
```diff
@@ -0,0 +1,790 @@
...
+        type: string
+        format: iso-date-time
+        example: '2019-01-14T23:15:53.7270539Z'
+      updatedAt:
+        description: The time at which a given alarm instance was last updated
+        type: string
+        format: iso-date-time
+        example: '2019-01-14T23:15:53.7270539Z'
+      createdBy:
+        description: String identifying who or what created a given alarm instance. This is usually used to
```
caleyburton: Maybe "An identifier for who or what created a given alarm instance."?

alarm/nialarm.yml
```diff
@@ -0,0 +1,790 @@
...
+      displayName:
+        description: Optional display name for a given alarm instance. Display names do not need to be unique.
+        type: string
+        example: Low disk space on tester-1.
+      description:
+        description: Optional description text for a given alarm instance
+        type: string
+        example: Disk space on tester-1 is high. Disk usage was 95% when this alarm was created.
+      keywords:
+        description: Keywords associated with an alarm instance. Alarm instances can be tagged with keywords
```
caleyburton: Instead of stating the parameter's name in the first sentence of the description, could we say "Words or phrases" instead?

alarm/nialarm.yml
```diff
@@ -0,0 +1,790 @@
...
+          - keyword2
+      notes:
+        description: A collection of notes for a given alarm instance. Notes are set by humans to record
+          information about an alarm after it has been investigated.
+        type: array
+        items:
+          $ref: '#/definitions/AlarmNote'
+      properties:
+        description: Key-value-pair metadata associated with an alarm instance. Alarm instances can be tagged
+          with properties so that they can be more easily found with queries. Additionally, alarm instance
```
caleyburton: "Alarm instances can be tagged with properties to make it easier to find them with queries."

alarm/nialarm.yml
```diff
@@ -0,0 +1,790 @@
...
+          type: string
+        example:
+          key1: value1
+      resourceType:
+        description: The type of resource associated with a given alarm instance
+        type: string
+        example: Tag
+  AlarmNote:
+    title: Alarm Note
+    description: Contains information about a particular alarm instance, such as a description of an
```
caleyburton: Can we remove "contains" for consistency?

alarm/nialarm.yml
```diff
@@ -0,0 +1,790 @@
...
+      shortText:
+        description: Short description of the condition
+        type: string
+        example: Low disk space
+      detailText:
+        description: Detailed description of the condition
+        type: string
+        example: Disk usage on CRIO1 is 95% (lower than configured threshold of 90%)
+      keywords:
+        description: Keywords associated with a transition. Useful for attaching additional metadata to
```
caleyburton: "Words or phrases" instead of "Keywords"?

alarm/nialarm.yml
```diff
@@ -0,0 +1,790 @@
...
+        example: 4
+      createdBy:
+        type: string
+        description: Created by query. The service will return instances whose *createdBy* fields are equal to
+          the query, taking case into account.
+        example: TagRuleEngine
+      channel:
+        type: string
+        description: Channel query. The service will return instances whose *channel* fields are equal to the
+          query, taking case into account. Supports wildcard match checking uses glob-style wildcards.
```
caleyburton: "with" instead of "uses"?

alarm/nialarm.yml
```diff
@@ -0,0 +1,790 @@
...
+        type: object
+        description: Property query. The service will return instances whose *properties* fields contain all
+          of the specified key/value pairs.
+        additionalProperties:
+          type: string
+        example:
+          key1: value1
+      keywords:
+        type: array
+        description: Keyword query. The service will return instances whose *keywords* fields contain all of
```
caleyburton, line 391: "Words or phrases" instead of "Keyword"?

alarm/nialarm.yml
```diff
@@ -0,0 +1,790 @@
...
+      properties:
+        subQueries:
+          description: Additional queries. The service will return instances which match any of the queries
+            in the subQueries array. If a top-level query was provided, the top-level query will be AND'ed
+            with each of the sub queries.
+          type: array
+          items:
+            $ref: '#/definitions/SubQuery'
+        returnMostRecentlyOccurredOnly:
+          description: For the instances matched by this query, only return the most recent instance for each
```
caleyburton: "Instances matched by this query will return only the most recent instance for each alarmId."

alarm/nialarm.yml
```diff
@@ -0,0 +1,790 @@
...
+      tags: [alarm instances]
+      summary: Create or update an alarm instance
+      description: Creates or updates an instance of an alarm based on the requested transition and the state
+        of the current active alarm instance with the given *alarmId*.
+      operationId: CreateOrUpdateInstance
+      x-ni-operation: writeAlarmInstances
+      parameters:
+        - in: body
+          name: Request body
+          description: Request body. For requests which do not result in a new alarm instance being created,
```
caleyburton: In order to keep from restating the name in the description, could we update it to something like "Container which holds data for the request."?

alarm/nialarm.yml
```diff
@@ -0,0 +1,790 @@
...
+            type: object
+            title: CreateOrUpdateInstanceRequest
+            required:
+              - transition
+              - alarmId
+            properties:
+              alarmId:
+                description: A value meant to uniquely identify the particular process or condition tracked
+                  by a given alarm. For example, alarms created by the Tag Rule Engine Service have their
+                  alarmIds set to the concatenation of the path of the tag which caused the rule to be
```
caleyburton, line 601: Same question here as above: could "tag path" work here?

alarm/nialarm.yml
```diff
@@ -0,0 +1,790 @@
...
+                  triggered and the ID of the rule.
+                type: string
+                example: CRIO1.System.Health.DiskSpaceMeanUsePercentage.Bdd9u!4aMd!$pYrf*CnaIZ2tbu$-Ct%?
+              transition:
+                $ref: '#/definitions/AlarmTransition'
+              createdBy:
+                description: String identifying who or what created a given alarm instance. This is usually
```
caleyburton: Same comment as above. "An identifier for..."

alarm/nialarm.yml
```diff
@@ -0,0 +1,790 @@
...
+                description: Optional display name for a given alarm instance. Display names do not need to be
+                  unique.
+                type: string
+                example: Low disk space on tester-1.
+              description:
+                description: Optional description text for a given alarm instance
+                type: string
+                example: Disk space on tester-1 is high. Disk usage was 95% when this alarm was created.
+              keywords:
+                description: Keywords associated with an alarm instance. Alarm instances can be tagged with
```
caleyburton: "Words and phrases" instead of "keyword"?

alarm/nialarm.yml
```diff
@@ -0,0 +1,790 @@
...
+                  keywords to make it easier to find them with queries.
+                type: array
+                items:
+                  type: string
+                example:
+                  - keyword1
+                  - keyword2
+              properties:
+                description: Key-value-pair metadata associated with an alarm instance. Alarm instances can be
+                  tagged with properties so that they can be more easily found with queries. Additionally,
```
caleyburton: "Alarm instances can be tagged with properties to make it easier to find them with queries."

caleyburton (commented):
Hey Phil! I've gone through and added some comments with suggested rephrasing here and there. If you feel like a suggestion wouldn't be technically accurate and want me to keep working on it, please feel free to let me know. :)

alarm/nialarm.yml
pvallone: In this case I kind of like using the word "Keyword", as that is the field of the alarm instance which we are querying against.

alarm/nialarm.yml
pvallone: Yikes @ uses in my original string. Went with 'with'.

alarm/nialarm.yml
pvallone: Yep, did so.

alarm/nialarm.yml
pvallone: Yes, I like that much more.

alarm/nialarm.yml
pvallone: Nice, thanks for looking out for consistency here. I made changes to all of those strings.

alarm/nialarm.yml
pvallone: :)

alarm/nialarm.yml
pvallone: Yep!

alarm/nialarm.yml
pvallone: So it isn't really the tag's path which caused the rule to be triggered, it's the fact that the tag's value is within the set point. The path of the tag uniquely identifies the tag, so that's what we append to the alarmId.

alarm/nialarm.yml
pvallone: Definitely the latter.

alarm/nialarm.yml
pvallone: Yep, and I went with the latter, as the service refers to these things as alarm instances.

alarm/nialarm.yml
pvallone: It seems like this suggests that its the instances which are doing the returning, which isn't accurate. Reading my own string again, I wasn't really happy with it I revised it to the string below. Let me know if you have any suggestions. 

Whether or not the alarm service should group the instances matched by this query by *alarmId*, and then only return the most recent instance for each grouping. In this context, recency is based on when the alarm instance occurred, not when it was last updated.

alarm/nialarm.yml
pvallone: Yep, did so

alarm/nialarm.yml
pvallone: See my reply to the comment above.

caleyburton (approved):

spanglerco (approved):

PR #42, opened 2019-03-06
Add tag rule engine swagger doc

pvallone:

This PR adds a swagger doc for the Tag Rule Engine.

This PR is part of F21323, which is green for SystemLink 19.0.

- Verified that there were no errors in editor.swagger.io. 
- Tested the routes and their examples against SystemLink server.

### DIFF
--- a/tag-rule/nitagrule.yml
+++ b/tag-rule/nitagrule.yml
@@ -2,7 +2,7 @@ swagger: '2.0'
 info:
   version: '1'
   title: Tag Rule Engine Web Service
-  description: SystemLink Tag Rule Engine Service HTTP API
+  description: SystemLink Tag Rule Engine Service HTTP API. Only available on SystemLink Server.
   contact:
     name: National Instruments
     url: 'https://www.ni.com/systemlink'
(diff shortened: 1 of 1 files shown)


tag-rule/nitagrule.yml
```diff
@@ -0,0 +1,564 @@
...
+
+          - writeRules: The ability to create and modify tag alarm rules
+
+        type: object
+        properties:
+          deleteAlarmInstances:
+            $ref: '#/definitions/Operation'
+          readAlarmInstances:
+            $ref: '#/definitions/Operation'
+          writeAlarmInstances:
```
spanglerco: Are these the correct names? They don't match the description.

tag-rule/nitagrule.yml
```diff
@@ -0,0 +1,564 @@
...
+          schema:
+            description: Version information
+            title: RootEndpointResponse
+            type: object
+            properties:
+              v1:
+                $ref: '#/definitions/V1Operations'
+              version:
+                description: Implementation version of the web service
+                type: string
```
spanglerco, line 255: Should we provide an example here?

tag-rule/nitagrule.yml
```diff
@@ -0,0 +1,564 @@
...
+          condition with a set point of 10, a comparator of GREATER_THAN, and a deadband of 2 will request 
+          that alarms be cleared if the tag's values is less than or equal to (10 - 2) = 8. Similarly, if the
+          condition had a comparator of LESS_THAN, the condition would request that alarms be cleared if the
+          tag's value is greater than or equal to (10 + 2) = 12.
+        type: string
+        example: '2'
+      severityLevel:
+        description: An integer in the range [1,4] which specifies the severity level of a condition.
+          Higher numeric values correspond to higher severity levels.
+        type: number
```
spanglerco: `type: integer` would be more correct.

tag-rule/nitagrule.yml
```diff
@@ -0,0 +1,564 @@
...
+        maximum: 4
+        example: 2
+        default: 2
+      notificationStrategyIds:
+        description: The IDs of the notification strategies which should be triggered if a condition
+          results in alarms being created or increasing in severity.
+        type: array
+        items:
+          type: string
+        example: []
```
spanglerco, line 128: An empty list doesn't seem like a very helpful example :)

tag-rule/nitagrule.yml
```diff
@@ -0,0 +1,564 @@
...
+    description: The data type of the tags which a rule applies to. Rules are only evaluated
+      against tags whose data type matches their *tagDataType*.
+    type: string
+    enum: [DOUBLE, INT, STRING, BOOLEAN, U_INT64, DATE_TIME]
+    example: DOUBLE
+  SearchPath:
+    title: Search Path
+    description: A string describing which tag(s) a rule applies to. This can take the form of either
+      a single tag's path, or a string containing wildcard characters which can match multiple tags.
+    type: string
```
spanglerco: Codegen doesn't handle non-enum string definitions very well. For example, the C# client generator will give you a class like this:

`public partial class SearchPath :  IEquatable<SearchPath>, IValidatableObject`

with no properties, so you can't actually get or set anything. To make this easier to use for clients, we should consider inlining all the scalar definitions that aren't enums.

tag-rule/nitagrule.yml
```diff
@@ -0,0 +1,564 @@
...
+    example: '*.System.Health.DiskUsePercentage'
+  Conditions:
+    description: Array containing one or more conditions which control when a rule creates or clears alarms.
+    type: array
```
spanglerco, line 138: This one at least generates a class that works (at least for C#): `public partial class Conditions : List<Condition>,  IEquatable<Conditions>, IValidatableObject`, so you probably don't have to inline it.

tag-rule/nitagrule.yml
```diff
@@ -0,0 +1,564 @@
...
+            title: CreateRuleResponse
+            properties:
+              id:
+                description: The ID of the created rule
+                type: string
+                example: 5c40ec55e0d6441168b4c543
+        401:
+          description: Unauthorized
+          schema:
+            $ref: '#/definitions/Error'
```
spanglerco: Should the 401s be using the `#/responses/Unauthorized` ref?

tag-rule/nitagrule.yml
```diff
@@ -0,0 +1,564 @@
...
+      responses:
+        204:
+          description: No Content
+        401:
+          description: Unauthorized
+          schema:
+            $ref: '#/definitions/Error'
+        default:
+          $ref: '#/responses/Error'
+    patch:
```
spanglerco, line 392: Usually for patch, we specify the format we're using:

```yaml
consumes:
  - application/merge-patch+json
```

tag-rule/nitagrule.yml
```diff
@@ -0,0 +1,564 @@
...
+      tags: [rules]
+      summary: Delete rules
+      description: Deletes multiple rules.
+      operationId: DeleteRules
+      x-ni-operation: deleteRules
+      parameters:
+        - in: body
+          name: Request body
+          required: true
+          schema:
```
spanglerco, line 424: This schema needs a title for codegen.

tag-rule/nitagrule.yml
```diff
@@ -0,0 +1,564 @@
...
+      tags: [rules]
+      summary: Toggle rules
+      description: Enables or disables multiple rules.
+      operationId: ToggleRules
+      x-ni-operation: writeRules
+      parameters:
+        - in: body
+          name: Request body
+          required: true
+          schema:
```
spanglerco, line 454: This schema needs a title for codegen.

tag-rule/nitagrule.yml
```diff
@@ -0,0 +1,564 @@
...
+                items:
+                  type: string
+                example: [5c40ec55e0d6441168b4c543, 5c40ec55e0d6441168b4c543]
+              disabled:
+                $ref: '#/definitions/Disabled'
+      responses:
+        200:
+          description: OK
+          schema:
+            properties:
```
spanglerco, line 478: You should explicitly specify `type: object` to avoid codegen issues, as well as a title.

tag-rule/nitagrule.yml
```diff
@@ -0,0 +1,564 @@
...
+          required: true
+          schema:
+            properties:
+              ids:
+                description: The IDs of the rules which should be toggled.
+                type: array
+                items:
+                  type: string
+                example: [5c40ec55e0d6441168b4c543, 5c40ec55e0d6441168b4c543]
+              disabled:
```
spanglerco, line 467: Should either of these be marked as required?

tag-rule/nitagrule.yml
```diff
@@ -0,0 +1,564 @@
...
+      description: Deletes multiple rules.
+      operationId: DeleteRules
+      x-ni-operation: deleteRules
+      parameters:
+        - in: body
+          name: Request body
+          required: true
+          schema:
+            properties:
+              ids:
```
spanglerco, line 430: Should this be marked as required?

tag-rule/nitagrule.yml
```diff
@@ -0,0 +1,564 @@
...
+      summary: Query rules
+      description: Queries for rules.
+      operationId: QueryRules
+      x-ni-operation: readRules
+      parameters:
+        - in: body
+          name: Request body
+          required: true
+          schema:
+            properties:
```
spanglerco, line 509: This schema needs a title for codegen.

tag-rule/nitagrule.yml
```diff
@@ -0,0 +1,564 @@
...
+                example: low disk space
+              description:
+                description: Description query. Returns rules whose *description* fields contain the query
+                  string, without taking case into account.
+                type: string
+                example: turbine 5
+              tagPath:
+                description: Tag path query. If the query string contains wildcard characters, the service
+                  will return rules whose *searchPaths* either intersect with, or are matched by, the query
+                  string. For example, if the query string is CRIO1 .System.Health.*, then the service would
```
spanglerco: Is that an extra space after CRIO1?

tag-rule/nitagrule.yml
```diff
@@ -0,0 +1,564 @@
...
+                description: Number of entries to include in the response list. Typically used in combination
+                  with the skip parameter to implement pagination.
+                type: integer
+                default: 10000
+                maximum: 10000
+      responses:
+        200:
+          description: OK
+          schema:
+            properties:
```
spanglerco, line 566: Ditto on `type: object` and title for codegen.

spanglerco (changes requested):

tag-rule/nitagrule.yml
pvallone: Yep!

tag-rule/nitagrule.yml
pvallone: I left the non-string definitions around.

tag-rule/nitagrule.yml
pvallone: Inlined all of the string definitions

tag-rule/nitagrule.yml
pvallone: No, but it's intentionally empty. You can currently create a rule with a bogus notification strategy ID without errors, but I want the example to create valid rules which aren't going to log errors when they're evaluated. We might also change the service in the future to validate that the notification strategies are valid at rule creation time.

tag-rule/nitagrule.yml
pvallone: Nope...

tag-rule/nitagrule.yml
pvallone: Don't think I did so for alarm, but I added "18.0"

tag-rule/nitagrule.yml
pvallone: Yep, both are required and are now marked as such.

tag-rule/nitagrule.yml
pvallone: Yep, got rid of it

tag-rule/nitagrule.yml
```diff
@@ -344,7 +353,7 @@ paths:
         401:
           description: Unauthorized
           schema:
-            $ref: '#/definitions/Error'
+            $ref: '#/responses/Unauthorized'
```
spanglerco: Responses refs have to go one more level up, not under a schema. Applies to below places as well.

spanglerco (changes requested):

tag-rule/nitagrule.yml
spanglerco: Makes sense

tag-rule/nitagrule.yml
pvallone: They should all be cleaned up now

spanglerco (dismissed):

tag-rule/nitagrule.yml
```diff
@@ -0,0 +1,579 @@
...
+  Condition:
+    title: Condition
+    description: An object describing the circumstances under which an alarm should be created.
+    type: object
+    required:
+      - setPoints
+      - comparator
+    properties:
+      setPoints:
+        description: Values which the tag's current value will be compared to when determining if the
```
caleyburton: What if we condensed this description a little? "Values compared against the tag's current value to determine whether or not the condition is true." Does this feel accurate still?

tag-rule/nitagrule.yml
```diff
@@ -0,0 +1,579 @@
...
+    type: string
+    enum: [DOUBLE, INT, STRING, BOOLEAN, U_INT64, DATE_TIME]
+    example: DOUBLE
+  Conditions:
+    description: Array containing one or more conditions which control when a rule creates or clears alarms.
+    type: array
+    items:
+      $ref: '#/definitions/Condition'
+  Keywords:
+    description: Words or phrases associated with a rule. Rules can be taggedwith keywords to make it easier
```
caleyburton: "tagged with" (missing space)

tag-rule/nitagrule.yml
```diff
@@ -0,0 +1,579 @@
...
+    description: Key-value-pair metadata associated with a rule. Rules can be tagged with properties to make
+      it easier to find them with queries.
+    type: object
+    additionalProperties:
+      type: string
+    example:
+      key1: value1
+  RuleCommonFields:
+    title: Rule Common Fields
+    description: Type single-sourcing the definition of the fields available in both an update request and in
```
caleyburton: "A single-sourced definition of available fields in both an update request and the rule type itself"?

tag-rule/nitagrule.yml
```diff
@@ -0,0 +1,579 @@
...
+        example: false
+        default: false
+      displayName:
+        description: Display string for a rule's name.
+        type: string
+        example: Low Disk Space Rule
+      description:
+        description: Display string for a rule's description.
+        type: string
+        example: Preconfigured rule for monitoring managed systems' disk space
```
caleyburton: Should we use "system's" instead? While the rule might be applied all of the managed systems, the user is likely going to see each system's disk space value separately, right?

tag-rule/nitagrule.yml
```diff
@@ -0,0 +1,579 @@
...
+                $ref: '#/definitions/Properties'
+      responses:
+        200:
+          description: OK
+          schema:
+            type: object
+            title: CreateRuleResponse
+            properties:
+              id:
+                description: The ID of the created rule
```
caleyburton: Should we mention whether it's unique or not here?

caleyburton (changes requested):

tag-rule/nitagrule.yml
pvallone: Sure, that sounds good to me. Did you want me to drop the two sentences after it?

> Conditions associated with the IN_RANGE and OUT_OF_RANGE comparators will have two set points. All other conditions should have only one set point.

tag-rule/nitagrule.yml
pvallone: Correct - the rule will be applied to all managed systems, but separate alarms will be created for each system. I wasn't sure what grammar dictated here, but "system's" works for me!

tag-rule/nitagrule.yml
pvallone: I think most users will assume that IDs are unique, but it's OK to call it out here to dispel confusion. We also mention that it's unique on the definition of the ID field on the rule type.

caleyburton (approved):

spanglerco (approved):

PR #45, opened 2019-03-18
Add Swagger document for tag historian

spanglerco:

Documents the existing tag historian HTTP API.

Provides documentation for developing clients for the tag historian API.

Imported the document into Insomnia and made several requests. Generated a C# client and verified it built and appeared to have all of the correct data types.

### DIFF
--- a/tag-historian/nitaghistorian.yml
+++ b/tag-historian/nitaghistorian.yml
@@ -42,7 +42,7 @@ definitions:
       message: 'Invalid id: 5c88d2ae26b41e22f08b4407.'
       args: []
   ErrorEntry:
-    description: Populated with any errors that ocurr when performing multiple operations in a single
+    description: Populated with any errors that occur when performing multiple operations in a single
       request, with one inner error per failed operation.
     title: Error Entry
     properties:
@@ -128,7 +128,7 @@ definitions:
         type: string
         example: '5c88d2ae26b41e22f08b4407'
       path:
-        description: The path of the tag corresponding to this historical value.
+        description: The path of the tag corresponding to the historical value.
         type: string
         example: system1.tag1
       type:
@@ -140,7 +140,7 @@ definitions:
         type: string
         example: '12.87'
       timestamp:
-        description: The ISO-8601 date and time that the historical value was written, in UTC.
+        description: The ISO-8601 formatted timestamp when the historical value was written, in UTC.
         type: string
         format: iso-date-time
         example: '2019-03-13T09:51:38Z'
@@ -154,8 +154,8 @@ definitions:
           note: This value is an outlier and needs investigation.
       expirationDate:
         description: For historical values of tags that aren't using a count-based retention policy,
-          the expiration date is the date and time at which the historical value will be deleted
-          automatically.
+          the expiration date is the ISO-8601 formatted timestamp at which the historical value will
+          be deleted automatically.
         type: string
         format: iso-date-time
         example: '2020-03-13T09:51:38Z'
@@ -165,7 +165,7 @@ definitions:
     title: Http Historical Value Update
     properties:
       id:
-        description: The ID of the historical value to update.
+        description: The identifier of the historical value to update.
         type: string
         example: '5c88d2ae26b41e22f08b4407'
       properties:
@@ -176,8 +176,8 @@ definitions:
         example:
           note: This value is an outlier and needs investigation.
       expirationDate:
-        description: The expiration date is the date and time at which the historical value will be
-          deleted automatically, in ISO-8601 format. Do not specify this value for tags that use
+        description: The expiration date is the ISO-8601 formatted timestamp at which the historical
+          value will be deleted automatically. Do not specify this value for tags that use
           count-based retention.
         type: string
         format: iso-date-time
@@ -210,7 +210,7 @@ definitions:
           - system1.tag1
           - system2.tag1
       startTime:
-        description: Limits the returned historical values to those written at or after at the given
+        description: Limits the returned historical values to those written at or after the given
           time in ISO-6801 format.
         type: string
         format: iso-date-time
@@ -244,8 +244,8 @@ definitions:
         minimum: 0
         default: 0
       sortOrder:
-        description: Specifies the order to return historical values, either ascending or descending
-          order based on when each tag value was written.
+        description: Specifies ascending or descending order to return historical values, based
+          on when each tag was written.
         type: string
         enum: [ASCENDING, DESCENDING]
         default: ASCENDING
@@ -262,7 +262,7 @@ definitions:
         type: string
         example: system1.tag1
       type:
-        description: The value's data type.
+        description: The data type of the value.
         type: string
         enum: [DOUBLE, INT, STRING, BOOLEAN, U_INT64, DATE_TIME]
       value:
@@ -270,7 +270,7 @@ definitions:
         type: string
         example: '12.87'
       timestamp:
-        description: The ISO-8601 date and time that the historical value was written, in UTC.
+        description: The ISO-8601 formatted timestamp that the historical value was written, in UTC.
         type: string
         format: iso-date-time
         example: '2019-03-13T09:51:38Z'
@@ -284,8 +284,8 @@ definitions:
           note: This value is an outlier and needs investigation.
       expirationDate:
         description: For historical values of tags that aren't using a count-based retention policy,
-          the expiration date is the date and time at which the historical value will be deleted
-          automatically.
+          the expiration date is the ISO-8601 formatted timestamp at which the historical value will
+          be deleted automatically.
         type: string
         format: iso-date-time
         example: '2020-03-13T09:51:38Z'
@@ -333,7 +333,7 @@ definitions:
       query:
         $ref: '#/definitions/HistoricalTagQuery'
       requestId:
-        description: The ID of the asynchronous query to associate with a result in a later message.
+        description: The identifier of the asynchronous query to associate with a result in a later message.
         type: string
         example: '4cbba25a-c1b8-4fbe-8829-9ddfde555188'
       hasError:
@@ -383,7 +383,7 @@ definitions:
         enum: ['TagHistorianAsyncQueryRequest']
         example: TagHistorianAsyncQueryRequest
       queries:
-        description: The list of history queries to perform.
+        description: The list of historical queries to perform.
         type: array
         items:
           $ref: '#/definitions/HistoricalTagQuery'
@@ -408,7 +408,7 @@ definitions:
         enum: ['TagHistorianAsyncQueryResponse']
         example: TagHistorianAsyncQueryResponse
       queryJobs:
-        description: A list containing the information about each started query.
+        description: A list containing the information about each query started.
         type: array
         items:
           $ref: '#/definitions/HistoricalTagAsyncQueryJob'
@@ -427,7 +427,7 @@ definitions:
             enum: ['TagHistorianAsyncQueryDataRoutedMessage']
             example: TagHistorianAsyncQueryDataRoutedMessage
           requestId:
-            description: The ID of the completed asynchronous query as given in the corresponding
+            description: The identifier of the completed asynchronous query as given in the corresponding
               *WebSocketAsyncQueryResponse* message.
             type: string
             example: '4cbba25a-c1b8-4fbe-8829-9ddfde555188'
@@ -444,7 +444,7 @@ definitions:
         enum: ['TagHistorianAsyncQueryCompleteRoutedMessage']
         example: TagHistorianAsyncQueryCompleteRoutedMessage
       queryJobs:
-        description: A list containing the information about each started query.
+        description: A list containing the information about each query started.
         type: array
         items:
           $ref: '#/definitions/HistoricalTagAsyncQueryJob'
@@ -783,14 +783,14 @@ paths:
         - in: body
           name: Request body
           required: true
-          description: Contains the IDs of the historical values to delete.
+          description: Contains the identifiers of the historical values to delete.
           schema:
-            description: Contains the IDs of the historical values to delete.
+            description: Contains the identifiers of the historical values to delete.
             type: object
             title: Tag Historian Delete Values By Id Request
             properties:
               ids:
-                description: The IDs of the historical values to delete.
+                description: The identifiers of the historical values to delete.
                 type: array
                 items:
                   type: string
@@ -821,7 +821,7 @@ paths:
           name: id
           type: string
           required: true
-          description: The ID of the historical value to retrieve.
+          description: The identifier of the historical value to retrieve.
           x-example: '5c88d2ae26b41e22f08b4407'
       responses:
         200:
@@ -854,8 +854,8 @@ paths:
           name: id
           type: string
           required: true
-          description: The ID of the historical value to modify. This value is used in preference to
-            the ID contained in the request body.
+          description: The identifier of the historical value to modify. This value is used in preference to
+            the identifier contained in the request body.
           x-example: '5c88d2ae26b41e22f08b4407'
         - in: body
           name: Request body
@@ -897,7 +897,7 @@ paths:
           name: id
           type: string
           required: true
-          description: The ID of the historical value to delete.
+          description: The identifier of the historical value to delete.
           x-example: '5c88d2ae26b41e22f08b4407'
       responses:
         204:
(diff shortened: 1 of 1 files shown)


tag-historian/nitaghistorian.yml
```diff
@@ -0,0 +1,928 @@
+swagger: '2.0'
+info:
+  version: '1'
+  title: Tag Historian Web Service
+  description: SystemLink Tag Historian Service HTTP API.
```
pvallone, line 5: I was going to suggest adding "Only available on SystemLink Server", but I guess the SLC folks are working on that for this release.

tag-historian/nitaghistorian.yml
```diff
@@ -0,0 +1,928 @@
...
+        additionalProperties:
+          type: string
+        example:
+          note: This value is an outlier and needs investigation.
+      expirationDate:
+        description: The expiration date is the date and time at which the historical value will be deleted automatically, in ISO-8601 format. Do not specify this value for tags that use count-based retention.
+        type: string
+        format: iso-date-time
+        example: '2020-03-13T09:51:38Z'
+    required: [id]
```
pvallone, line 185: slick `required` syntax

tag-historian/nitaghistorian.yml
```diff
@@ -0,0 +1,928 @@
...
+    type: object
+    title: Tag Historian Values
+    properties:
+      values:
+        description: An object containing the historical value for the requested tag if it was successfully queried. The object's property name is the tag path.
+        additionalProperties:
+          type: array
+          items:
+            $ref: '#/definitions/HistoricalValue'
+        example:
```
pvallone, line 304: I think it'd be helpful if the example had three tag values in it, with two of them being associated with the same tag. Looking at the example response, I don't see any arrays, and I expected to see one.

tag-historian/nitaghistorian.yml
```diff
@@ -0,0 +1,928 @@
...
+      requestId:
+        description: The ID of the asynchronous query to associate with a result in a later message.
+        type: string
+        example: '4cbba25a-c1b8-4fbe-8829-9ddfde555188'
+      hasError:
+        description: Whether the query had an error and cannot be completed.
+        type: boolean
+  WebSocketQueryRequest:
+    title: Web Socket Query Request
+    description: A web socket message send by the client to query for historical tag values. The server will respond with a *WebSocketQueryResponse* message once the query has been completed.
```
pvallone: send -> sent

tag-historian/nitaghistorian.yml
```diff
@@ -0,0 +1,928 @@
...
+        properties:
+          action:
+            description: The action to perform. Must be set to "TagHistorianQueryRequest".
+            type: string
+            enum: ['TagHistorianQueryRequest']
+            example: TagHistorianQueryRequest
+      - $ref: '#/definitions/HistoricalTagQuery'
+  WebSocketQueryResponse:
+    title: Web Socket Query Response
+    description: A web socket message send by the server in response to a *WebSocketQueryRequest* containing the result of the query.
```
pvallone: send -> sent

tag-historian/nitaghistorian.yml
```diff
@@ -0,0 +1,928 @@
...
+      query:
+        $ref: '#/definitions/HistoricalTagQuery'
+      requestId:
+        description: The ID of the asynchronous query to associate with a result in a later message.
+        type: string
+        example: '4cbba25a-c1b8-4fbe-8829-9ddfde555188'
+      hasError:
+        description: Whether the query had an error and cannot be completed.
+        type: boolean
+  WebSocketQueryRequest:
```
pvallone, line 342: Does the WebSocket stuff work? If not, we may not want to publicize it just yet.

tag-historian/nitaghistorian.yml
```diff
@@ -0,0 +1,928 @@
...
+      action:
+        description: The action to perform. Must be set to "TagHistorianAsyncSubscribeRequest".
+        type: string
+        enum: ['TagHistorianAsyncSubscribeRequest']
+        example: TagHistorianAsyncSubscribeRequest
+      searchPaths:
+        description: Contains one or more tag paths to subscribe to. The paths may contain wildcards to subscribe to similar tags.
+        type: array
+        items:
+          type: string
```
pvallone, line 473: Should we include some tag path-esque example strings?

tag-historian/nitaghistorian.yml
```diff
@@ -0,0 +1,928 @@
...
+                    value: '12.87'
+                    timestamp: '2019-03-13T09:51:38Z'
+                    properties:
+                      note: This value is an outlier and needs investigation.
+                    expirationDate: '2020-03-13T09:51:38Z'
+              totalCount:
+                description: The total number of historical values matched by the query regardless of the take parameter. When applying decimation, this value is always equal to the number of returned values.
+                type: integer
+                minimum: 0
+                example: 1100
```
pvallone, line 716: Might be nice for the example totalCount to match the number of values in the example response

tag-historian/nitaghistorian.yml
```diff
@@ -0,0 +1,928 @@
...
+              errors:
+                $ref: '#/definitions/ErrorEntry'
+              failed:
+                description: The paths of the tags that had errors.
```
pvallone, line 720: Should we reiterate that this is always null? Or is this field populated sometimes?

tag-historian/nitaghistorian.yml
```diff
@@ -0,0 +1,928 @@
...
+          description: Success
+        401:
+          $ref: '#/responses/Unauthorized'
+        default:
+          $ref: '#/responses/Error'
+  /v1/tags/flush-values:
+    post:
+      tags: [utility]
+      summary: Process buffered tag updates
+      description: Triggers the tag historian to immediately begin processing tag updates. Clients do not need to call this method for updates to be processed normally.
```
pvallone: Do you think it would be valuable to call out the default batch interval? It's probably something like every 5 minutes or 1000 updates, whichever comes first.

tag-historian/nitaghistorian.yml
```diff
@@ -0,0 +1,928 @@
...
+        enum: ['TagHistorianAsyncValuesProcessedRoutedMessage']
+        example: TagHistorianAsyncValuesProcessedRoutedMessage
+      values:
+        description: The newly recorded historical values.
+        type: array
+        items:
+          $ref: '#/definitions/HistoricalValue'
+  WebSocketUnsubscribeRequest:
+    title: Web Socket Unsubscribe Request
+    description: A web socket message sent by the client to unsubscribe from notifications when one or more tags have recorded new historical values created by a previous *WebSocketSubscribeRequest*. The server will respond with a *WebSocketUnsubscribeResponse* message to confirm.
```
pvallone: This string is a little long, but maybe that's just my laptop. I'd been setting a 110 character column guide for the swagger docs I've worked on so far.

tag-historian/nitaghistorian.yml
```diff
@@ -0,0 +1,928 @@
...
+          name: Request body
+          required: true
+          description: Defines how historical values are queried and for which tags.
+          schema:
+            description: Defines how historical values are queried and for which tags.
+            type: object
+            title: Http Query Request Body
+            properties:
+              paths:
+                description: One or more paths identifying the tags whose historical values should be returned.
```
pvallone: Are wildcards supported?

tag-historian/nitaghistorian.yml
spanglerco: Yeah, that's what I wasn't sure about either, so I left it out.

tag-historian/nitaghistorian.yml
spanglerco: Yeah, the example was missing the array. Though this is only used in a route that will always give 0 or 1 values, so the example shouldn't have more than one either.

tag-historian/nitaghistorian.yml
spanglerco: Yes, the tag viewer web UI uses it, though it only uses the async query and subscribe messages, not sync query or unsubscribe. So I assume SLC will need to implement it too, or they will need the web UI changed.

tag-historian/nitaghistorian.yml
spanglerco: Gah, missed some :)

tag-historian/nitaghistorian.yml
spanglerco: No, only tag paths.

tag-historian/nitaghistorian.yml
spanglerco: My thinking was the example request body has a take of 1 (now 2), so the response would only have that many elements with a larger totalCount.

tag-historian/nitaghistorian.yml
spanglerco: In this response it's sometimes populated. That was a different one where it's always null. Tag historian is fun :)

tag-historian/nitaghistorian.yml
spanglerco: Well, the route is really only meant for testing and not something we expect clients to ever call. Updates are something like 5 seconds, and aging is like an hour. I don't really want to try to pin down the exact numbers here.

tag-historian/nitaghistorian.yml
spanglerco: Yeah, the Swagger editor automatically wraps in the editor but not in the file, so it's easy to forget.

tag-historian/nitaghistorian.yml
pvallone: Ah, I'd missed the take.

tag-historian/nitaghistorian.yml
pvallone: Sounds reasonable.

pvallone (dismissed):

tag-historian/nitaghistorian.yml
```diff
@@ -0,0 +1,1019 @@
...
+        type: array
+        items:
+          type: string
+    example:
+      name: TagHistorian.InvalidId
+      code: -251913
+      message: 'Invalid id: 5c88d2ae26b41e22f08b4407.'
+      args: []
+  ErrorEntry:
+    description: Populated with any errors that ocurr when performing multiple operations in a single
```
caleyburton: Occur is misspelled.

tag-historian/nitaghistorian.yml
```diff
@@ -0,0 +1,1019 @@
...
+  HttpHistoricalValue:
+    type: object
+    title: Http Historical Value
+    properties:
+      id:
+        description: The unique identifier for referencing the historical value in other requests.
+        type: string
+        example: '5c88d2ae26b41e22f08b4407'
+      path:
+        description: The path of the tag corresponding to this historical value.
```
caleyburton: maybe swap out "this" for "the"?

tag-historian/nitaghistorian.yml
```diff
@@ -0,0 +1,1019 @@
...
+      type:
+        description: The value's data type.
+        type: string
+        enum: [DOUBLE, INT, STRING, BOOLEAN, U_INT64, DATE_TIME]
+      value:
+        description: The value of the tag at the time it was written, converted to a string.
+        type: string
+        example: '12.87'
+      timestamp:
+        description: The ISO-8601 date and time that the historical value was written, in UTC.
```
caleyburton: "The ISO-8601 formatted timestamp when the historical value was written in UTC."?

tag-historian/nitaghistorian.yml
```diff
@@ -0,0 +1,1019 @@
...
+        type: string
+        format: iso-date-time
+        example: '2020-03-13T09:51:38Z'
+  HttpHistoricalValueUpdate:
+    description: An update for a single historical value.
+    type: object
+    title: Http Historical Value Update
+    properties:
+      id:
+        description: The ID of the historical value to update.
```
caleyburton: Should we swap out "ID" for "identifier" here? If you agree, let's implement that throughout the doc.

tag-historian/nitaghistorian.yml
```diff
@@ -0,0 +1,1019 @@
...
+        description: Key-value pairs associated with the historical value after the fact. Tag
+          properties are not copied to the historical value.
+        type: object
+        additionalProperties:
+          type: string
+        example:
+          note: This value is an outlier and needs investigation.
+      expirationDate:
+        description: For historical values of tags that aren't using a count-based retention policy,
+          the expiration date is the date and time at which the historical value will be deleted
```
caleyburton: Should we state "ISO-8601 format" or "ISO-8601 formatted timestamp" instead of "date and time" instead? If you think the latter one, perhaps we should implement it throughout the doc?

tag-historian/nitaghistorian.yml
```diff
@@ -0,0 +1,1019 @@
...
+      properties:
+        description: Key-value pairs to associate with the historical value.
+        type: object
+        additionalProperties:
+          type: string
+        example:
+          note: This value is an outlier and needs investigation.
+      expirationDate:
+        description: The expiration date is the date and time at which the historical value will be
+          deleted automatically, in ISO-8601 format. Do not specify this value for tags that use
```
caleyburton: Whatever we decide for my comment above, we should replicate here. Let's remove the comma too.

tag-historian/nitaghistorian.yml
```diff
@@ -0,0 +1,1019 @@
...
+        default: 0
+        example: 1
+      decimation:
+        description: When non-zero, the returned historical values are decimated across requested time
+          span divided into the specified number of equal-sized intervals.
+        type: integer
+        minimum: 0
+        default: 0
+      sortOrder:
+        description: Specifies the order to return historical values, either ascending or descending
```
caleyburton: "The ascending or descending order to return historical values, based on when each tag was written."?

tag-historian/nitaghistorian.yml
```diff
@@ -0,0 +1,1019 @@
...
+      paths:
+        description: One or more paths identifying the tags whose historical values should be returned.
+        type: array
+        items:
+          type: string
+        example:
+          - system1.tag1
+          - system2.tag1
+      startTime:
+        description: Limits the returned historical values to those written at or after at the given
```
caleyburton: Let's remove the "at" following "after".

tag-historian/nitaghistorian.yml
```diff
@@ -0,0 +1,1019 @@
...
+      id:
+        description: The unique identifier of the retrieved historical value.
+        type: string
+        example: '5c88d2ae26b41e22f08b4407'
+      path:
+        description: The path of the tag corresponding to the historical value.
+        type: string
+        example: system1.tag1
+      type:
+        description: The value's data type.
```
caleyburton: "The data type of the value."?

tag-historian/nitaghistorian.yml
```diff
@@ -0,0 +1,1019 @@
...
+      - action
+      - queries
+    properties:
+      action:
+        description: The action to perform. Must be set to "TagHistorianAsyncQueryRequest".
+        type: string
+        enum: ['TagHistorianAsyncQueryRequest']
+        example: TagHistorianAsyncQueryRequest
+      queries:
+        description: The list of history queries to perform.
```
caleyburton: Should we state "historical" instead of "history" here?

tag-historian/nitaghistorian.yml
```diff
@@ -0,0 +1,1019 @@
...
+      and a *WebSocketAsyncQueryComplete* once all queries have completed.
+    type: object
+    properties:
+      action:
+        description: The action performed. Must be set to "TagHistorianAsyncQueryResponse".
+        type: string
+        enum: ['TagHistorianAsyncQueryResponse']
+        example: TagHistorianAsyncQueryResponse
+      queryJobs:
+        description: A list containing the information about each started query.
```
caleyburton: Could we flip "started query" to "query started"? If you agree, please implement throughout the doc.

caleyburton (commented):
Just some minor changes overall. Let me know if you want me to iterate on any of my suggestions or feel like a particular description needs a closer look.

caleyburton (changes requested):
Oops. Meant to select "request changes" vs. "comment."

tag-historian/nitaghistorian.yml
spanglerco: Switched to using `ISO-8601 formatted timestamp` everywhere instead of date and time.

caleyburton (approved):

PR #47, opened 2019-04-16
Webappservices swagger document

radu-iviniciu:
Adding SLC webappservices swagger document

### DIFF
--- a/webapp/webappservices.yaml
+++ b/webapp/webappservices.yaml
@@ -1,0 +1,301 @@
+swagger: '2.0'
+basePath: /niapp/v1
+info:
+  version: '1.0'
+  title: SystemLink Cloud Webapp Service 
+securityDefinitions:
+  ApiKeyAuth:
+    type: apiKey
+    name: x-ni-api-key
+    in: header
+security:
+  - ApiKeyAuth: []
+schemes:
+  - https
+paths:
+  '/webapps/{id}':
+    get:
+      tags:   
+        - Webapp metadata
+      summary: ''
+      responses:
+        '200':
+          description: A webapp record
+          schema: 
+            $ref: '#/definitions/WebappRecord'
+        '401':
+          description: API key is missing or invalid
+        '404':
+          description: Webapp with the given ID not found
+      parameters:
+        - in: path
+          name: id
+          required: true
+          description: The webapp's ID
+          type: string
+      description: Get webapp metadata by Id
+    put:
+      tags:
+        - Webapp metadata
+      summary: ''
+      responses:
+        '200':
+          description: Updated webapp metadata record
+          schema: 
+            $ref: '#/definitions/WebappRecord'
+        '401':
+          description: API key is missing or invalid
+        '404':
+          description: Webapp with the given ID not found
+        '400':
+          description: Ivalid webapp record
+      parameters:
+        - in: path
+          name: id
+          required: true
+          description: The webapp's ID
+          type: string
+        - in: body
+          required: true
+          name: webapp-binary-content
+          description: New webapp metadata
+          schema: 
+            $ref: '#/definitions/UpdateWebappMetadatReuqest'
+      operationId: update-webapp-metadata
+      description: Update webapp metadata by ID
+    delete:
+      tags:   
+        - Webapp metadata
+      summary: ''
+      responses:
+        '200':
+          description: Success
+        '401':
+          description: API key is missing or invalid
+        '404':
+          description: Webapp with the given ID not found
+      parameters:
+        - in: path
+          name: id
+          required: true
+          description: The webapp's ID
+          type: string
+      operationId: delete-webapp
+      description: Delete webapp by id
+  "/webapps":
+    get:
+      tags:   
+        - Webapp metadata
+      responses:
+        '200':
+          description: A list of webapp metadata records
+          schema: 
+            type: array 
+            items: 
+              $ref: '#/definitions/WebappRecord'
+        '401':
+          description: API key is missing or invalid
+      operationId: list-webapps
+      description: List users webapps
+    post:
+      tags:   
+        - Webapp metadata
+      responses:
+        '200':
+          description: Returns newly created webapp record 
+          schema: 
+            $ref: '#/definitions/WebappRecord'
+        '401':
+          description: API key is missing or invalid
+      operationId: create-webapp-record
+      description: Creates a new webapp record
+  "/webapps/shared-emails":
+    get:
+      tags:   
+        - Webapp metadata
+      responses:
+        '200':
+          description: List of emails the user has shared webapps with
+          schema: 
+            type: array 
+            items: 
+              type: string
+              example: john.doe@email.com
+        '401':
+          description: API key is missing or invalid
+      operationId: list-shared-emails
+      description: List the emails the user has shared webapps with
+  '/webapps/{id}/content':
+    get:
+      tags:
+        - Webapp content    
+      parameters:
+        - in: path
+          name: id
+          required: true
+          description: The webapp's ID
+          type: string
+      responses:
+        '200':
+          description: The webapp's content. JSON for Dashboards, HTML for WebVIs
+        '401':
+          description: API key is missing or invalid
+        '404':
+          description: Webapp with the given id not found
+      operationId: get-webapp-content
+      description: Get webapp content. ContentType varies from JSON for dashboards and templates or redirect to main HTML for WebVIs
+    put:
+      tags:
+        - Webapp content
+      summary: ''
+      responses:
+        '200':
+          description: Success
+        '401':
+          description: API key is missing or invalid
+        '404':
+          description: Webapp with the given ID not found
+        '400':
+          description: Invalid webapp content for webapp's type
+      parameters:
+        - in: path
+          name: id
+          required: true
+          description: The webapp's ID
+          type: string
+        - in: body
+          required: true
+          name: webapp-binary-content
+          description: The webapp content to upload
+          schema: 
+            $ref: '#/definitions/WebappContent'
+      operationId: update-webapp-content
+      description: Upload content for exisitng webapp
+  '/api/webapps/{id}/content/{path}':
+    get:
+      tags:
+        - Webapp content    
+      parameters:
+        - in: path
+          name: id
+          required: true
+          description: The webapp's ID
+          type: string
+        - in: path
+          name: path
+          required: true
+          type: string
+          description:  Path to inner webapp file 
+      responses:
+        '200':
+          description: The webapp's file. Content type can vary depending on what resource is (html, png etc.)
+          schema: 
+            type: string
+        '401':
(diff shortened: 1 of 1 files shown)


webapp/webappservices.yaml
```diff
@@ -0,0 +1,301 @@
...
+        type: array
+        items:
+          type: string
+        description: List of emails of users to share the webapp with. Applies when "shared" option is "direct"
+      policyIds: 
+        type: array
+        items:
+          type: string
+          example: "asdsad-17a6-45323-b64b-65325287372d"
+        description: List of policy IDs associated with the webapp giving it permissions to the user's SistemllinkCloud resources"
```
mollykreis, line 289: SistemllinkCloud -> SystemLink Cloud

This is done a few places throughout this document

webapp/webappservices.yaml
```diff
@@ -0,0 +1,301 @@
...
+          description: API key is missing or invalid
+        '404':
+          description: Webapp with the given ID not found
+      parameters:
+        - in: path
+          name: id
+          required: true
+          description: The webapp's ID
+          type: string
+      description: Get webapp metadata by Id
```
mollykreis, line 36: We should be consistent between ID and Id.

webapp/webappservices.yaml
```diff
@@ -0,0 +1,301 @@
...
+      operationId: duplicate-webapp-content
+      description: Duplicates webapp's content
+definitions:
+  WebappRecord:
+    type: object
+    title: Webapp record
+    properties:
+      id :
+        type: string
+        description: The webapp's
```
mollykreis, line 231: Looks like you're missing part of this description

webapp/webappservices.yaml
```diff
@@ -0,0 +1,301 @@
...
+        '200':
+          description: A list of webapp metadata records
+          schema: 
+            type: array 
+            items: 
+              $ref: '#/definitions/WebappRecord'
+        '401':
+          description: API key is missing or invalid
+      operationId: list-webapps
+      description: List users webapps
```
mollykreis, line 99: Lists the user's webapps

webapp/webappservices.yaml
```diff
@@ -0,0 +1,301 @@
...
+          description: Webapp with the given ID not found
+      parameters:
+        - in: path
+          name: id
+          required: true
+          description: The webapp's ID
+          type: string
+      operationId: delete-webapp
+      description: Delete webapp by id
+  "/webapps":
```
mollykreis, line 85: The query parameters for this route should be documented.

Same for any other route that has query parameters.

webapp/webappservices.yaml
```diff
@@ -0,0 +1,301 @@
...
+        '200':
+          description: Success
+        '401':
+          description: API key is missing or invalid
+        '404':
+          description: Webapp with the given sourceId not found
+      operationId: duplicate-webapp-content
+      description: Duplicates webapp's content
+definitions:
+  WebappRecord:
```
mollykreis, line 225: From using the API, it looks like this also has created and updated timestamps. They should be documented here.

webapp/webappservices.yaml
```diff
@@ -0,0 +1,301 @@
...
+        '200':
+          description: Updated webapp metadata record
+          schema: 
+            $ref: '#/definitions/WebappRecord'
+        '401':
+          description: API key is missing or invalid
+        '404':
+          description: Webapp with the given ID not found
+        '400':
+          description: Ivalid webapp record
```
mollykreis, line 51: Ivalid -> Invalid

webapp/webappservices.yaml
```diff
@@ -0,0 +1,301 @@
...
+        description: List of emails of users to share the webapp with. Applies when "shared" option is "direct"
+      policyIds: 
+        type: array
+        items:
+          type: string
+          example: "asdsad-17a6-45323-b64b-65325287372d"
+        description: List of policy IDs associated with the webapp giving it permissions to the user's SistemllinkCloud resources"  
+      properties:
+        type: object
+        description: A map of key value proeprties associated with the webapp
```
mollykreis, line 292: proeprties -> properties

webapp/webappservices.yaml
```diff
@@ -0,0 +1,301 @@
...
+      userMessage:
+        type: string
+        description: Custom message to embedd in the email notification sent out to newly added emails in the sharedEmails property when the shared property is direct
```
mollykreis, line 295: embedd -> embed

The formatting of the last part of this sentence is inconsistent with other places in this document:
_From line 283: Applies when "shared" option is "direct"_

webapp/webappservices.yaml
```diff
@@ -0,0 +1,301 @@
...
+      policyIds: 
+        type: array
+        items:
+          type: string
+          example: "asdsad-17a6-45323-b64b-65325287372d"
+        description: List of policy IDs associated with the webapp giving it permissions to the user's SistemllinkCloud resources"  
+      properties:
+        type: object
+        description: A map of key value proeprties associated with the webapp
+  UpdateWebappMetadatReuqest:
```
mollykreis, line 265: Metadat -> Metadata
Reuqest -> Request

mollykreis (commented):
I didn't look over this document completely, but I did notice a few things while reading the swagger document.

webapp/webappservices.yaml
```diff
@@ -0,0 +1,301 @@
...
+      parameters:
+        - in: path
+          name: id
+          required: true
+          description: The webapp's ID
+          type: string
+      description: Get webapp metadata by Id
+    put:
+      tags:
+        - Webapp metadata
```
alexweav, line 39: No need to prefix the sections with "webapp" since they're viewed in the context of the webapp service. They can just be `metadata` and `content` respectively.

webapp/webappservices.yaml
```diff
@@ -0,0 +1,301 @@
...
+security:
+  - ApiKeyAuth: []
+schemes:
+  - https
+paths:
+  '/webapps/{id}':
+    get:
+      tags:   
+        - Webapp metadata
+      summary: ''
```
alexweav, line 20: Can summaries be added to all routes? It appears as the top-level description of the route before the user expands it, and is really helpful when reading the documentation. It appears like in this screenshot:

image

Generally, the summary is shorter than the description, because the description only appears when the route is clicked on. Usually, the summary would be something like "Gets metadata for a webapp" where the description would be a full sentence with a bit more detail like "Gets metadata for a single webapp by ID."

webapp/webappservices.yaml
```diff
@@ -0,0 +1,301 @@
...
+          description: A list of webapp metadata records
+          schema: 
+            type: array 
+            items: 
+              $ref: '#/definitions/WebappRecord'
+        '401':
+          description: API key is missing or invalid
+      operationId: list-webapps
+      description: List users webapps
+    post:
```
alexweav, line 100: Seems odd for a POST route to have no parameters at all. Shouldn't there be a way to pass in webapp data when creating a new one?

webapp/webappservices.yaml
```diff
@@ -0,0 +1,301 @@
...
+        - Webapp metadata
+      summary: ''
+      responses:
+        '200':
+          description: Success
+        '401':
+          description: API key is missing or invalid
+        '404':
+          description: Webapp with the given ID not found
+      parameters:
```
alexweav, line 77: A bit of a nitpick, but it reads strangely to have parameters below responses, it feels like responses should come after. Feel free to ignore this one if you disagree, though.

webapp/webappservices.yaml
```diff
@@ -0,0 +1,301 @@
...
+          type: string
+        description: List of emails of users to share the webapp with. Applies when "shared" option is "direct"
+      policyIds: 
+        type: array
+        items:
+          type: string
+          example: "asdsad-17a6-45323-b64b-65325287372d"
+        description: List of policy IDs associated with the webapp giving it permissions to the user's SistemllinkCloud resources"  
+      properties:
+        type: object
```
alexweav, line 291: Usually, the pattern for this is something like:

```
      properties:
        description: Key-value pairs associated with the webapp.
        type: object
        additionalProperties:
          type: string
        example:
          key1: value1
```

That way, example values show up correctly. If you don't do the `additionalProperties` thing, the code generators in some languages will try to replace this with a generic object, rather than the string-map equivalent that is supposed to represent properties.

webapp/webappservices.yaml
```diff
@@ -0,0 +1,301 @@
...
+        description: A map of key value proeprties associated with the webapp
+      userMessage:
+        type: string
+        description: Custom message to embedd in the email notification sent out to newly added emails in the sharedEmails property when the shared property is direct
+        
+  WebappContent:
```
alexweav, line 297: Can you add an example here? If it shows up as base64 data, add a random base64 string. Otherwise, use a different placeholder string like "Webapp content." Right now, the example in documentation is just the word "string" which I don't think conveys the right message.

webapp/webappservices.yaml
```diff
@@ -0,0 +1,301 @@
...
+          description: Ivalid webapp record
+      parameters:
+        - in: path
+          name: id
+          required: true
+          description: The webapp's ID
+          type: string
+        - in: body
+          required: true
+          name: webapp-binary-content
```
alexweav, line 60: Metadata isn't binary content, right? This might need to be renamed.

webapp/webappservices.yaml
```diff
@@ -0,0 +1,301 @@
...
+        - in: path
+          type: string
+          required: true
+          name: sourceId
+          description: The id of the source webapp to get copy content from
+      responses:
+        '200':
+          description: Success
+        '401':
+          description: API key is missing or invalid
```
alexweav, line 219: Usually, we add a top-level `responses` section to the document which contains something like:

```
responses:
  Unauthorized:
    description: API key is missing or invalid
    headers:
      WWW_Authenticate:
        description: Information for how to authenticate
        type: string
```

Then, reference it like:

```
      responses:
        401:
```

This makes it show up a bit cleaner in the documentation:

image

(except it will have your description text instead of "Not Authorized - this was just the example I found)

webapp/webappservices.yaml
```diff
@@ -0,0 +1,301 @@
...
+        items:
+          type: string
+          example: john.doe@ni.com
+        description: List of emails of users to share the webapp with. Applies when "shared" option is "direct"
+      policyIds: 
+        type: array
+        items:
+          type: string
+          example: "asdsad-17a6-45323-b64b-65325287372d"
+        description: List of policy IDs associated with the webapp giving it permissions to the user's SistemllinkCloud resources"
```
alexweav, line 261: "webapp giving it permissions" -> "webapp, which give it access"

alexweav (changes requested):

tschmittni (commented):
I posted an update here: https://github.com/ni/systemlink-OpenAPI-documents/pull/69

PR #49, opened 2019-04-24
Swagger definition for Systems Management Service

chotea:
This is a refactoring for the current Systems Management HTTP API. We are writing a Service in C# so we can now define a Swagger. More details on what is changing can be found here: https://nitalk.jiveon.com/docs/DOC-500988

Signed-off-by: Cristian Hotea <cristian.hotea@ni.com>

### DIFF
--- a/systems-management/nisysmgmt.yml
+++ b/systems-management/nisysmgmt.yml
@@ -1033,7 +1033,7 @@ definitions:
                 - noarch
                 - any
                 - all
-        required: [feedUris]
+          required: [feedUris]
     required: [systemConfigurations]
 
   PackageResponse:
--- a/systems-management/nisysmgmt.yml
+++ b/systems-management/nisysmgmt.yml
@@ -1,6 +1,6 @@
 ﻿swagger: '2.0'
 info:
-  description: 'Management of distributed systems.'
+  description: 'Systems Management Service HTTP API. Manage and monitor your distributed systems.'
   version: '1'
   title: 'SystemLink Systems Management'
   contact:
@@ -53,7 +53,7 @@ responses:
         type: string
 
   PartialSuccess:
-    description: Partial Success in the case when only a part of the request body metadata information was processed successfully.
+    description: Partial Success indicates only part of the request body metadata information was processed successfully.
     schema:
       title: PartialSuccessResponse
       description: Partial success response containing error information.
@@ -74,7 +74,7 @@ responses:
 
 definitions:
   CreateJobRequest:
-    description: Request describing a job to be sent to systems.
+    description: Request describing a job to send to systems.
     title: Create Job Request
     type: object
     x-ni-cs-json-converter: JsonObjectConverter
@@ -95,7 +95,7 @@ definitions:
           - pkg.remove
           - pkg.install
       arg: 
-        description: An array of arrays of variable type elements that are arguments to the function specified by the "fun" property. If at least one function requires arguments, an array must be provided for each function, even if it is empty. Keyword arguments may be provided as a parameter of type object with at least one property of type boolean, "__kwarg__", that is evaluated to true.
+        description: An array of variable type element arrays, which are arguments to the function specified by the "fun" property. If at least one function requires arguments, an array must be provided for each function, even if it is empty. Keyword arguments may be provided as a parameter of type object with at least one property of type boolean ("__kwarg__") that evaluates to true.
         type: array
         items:
           type: array
@@ -114,7 +114,7 @@ definitions:
               pkgs:
               - acl-dbg: 1.0.0-0
       metadata:
-        description: Additional information passed along with the job.
+        description: Additional information of the job.
         type: object
         additionalProperties:
           type: object
@@ -127,7 +127,7 @@ definitions:
 
   CreateJobResponse:
     title: Create Job Response
-    description: Response describing a job that was sent to systems.
+    description: Response describing a job sent to systems.
     type: object
     properties:
       jid:
@@ -198,7 +198,7 @@ definitions:
     required: [systemsCredentials]
 
   SystemCredentials:
-    description: Credentials of a system.
+    description: The hostname, IP address, and login information of a system.
     title: System Credentials
     type: object
     properties:
@@ -221,7 +221,7 @@ definitions:
 
   RegisterSystemsResponse:
     title: Register Systems Response
-    description: Response of a request to register systems.
+    description: Response to a request to register systems to the server.
     type: object
     properties:
       requestId:
@@ -232,17 +232,17 @@ definitions:
 
   UnregisterSystemsRequest:
     title: Unregister Systems Request
-    description: Request to unregister systems.
+    description: Request to unregister systems from the server.
     type: object
     properties:
       tgt:
-        description: Array of string representing the IDs of systems to be removed.
+        description: Array of strings representing the IDs of systems to remove from SystemLink Server.
         type: array
         items:
           type: string
           example: VMware_Virtual_Platform--SN-VMware-56_4d_30_26_4f_c9_5c_05-b7_68_d1_cd_99_6e_22_16--MAC-02-01-02-03-04-08
       force:
-        description: Specify as true to remove systems from the database immediately, or false to wait until the unregister job returns from systems. If true, unregister job failures will not be cached.
+        description: A Boolean which specifies whether to remove systems from the database immediately (True) or wait until the unregister job returns from systems (False). If True, unregister job failures are not cached.
         type: boolean
         default: true
         example: false
@@ -278,7 +278,7 @@ definitions:
               $ref: '#/definitions/Error'
 
   CancelJobRequest:
-    description: Object containing the job ID and a system ID.
+    description: Object containing the job ID and system ID.
     title: Cancel Job Request
     type: object
     properties:
@@ -287,7 +287,7 @@ definitions:
         type: string
         example: '20190211135234448026_28320'
       systemId:
-        description: The system ID on which the job will be canceled.
+        description: The system ID on which the job is canceled.
         type: string
         example: VMware_Virtual_Platform--SN-VMware-56_4d_d2_08_ce_89_1b_70-e6_32_fc_21_6c_d0_d9_f9--MAC-00-0C-29-D0-D9-F9
     required: [jid, systemId]
@@ -339,19 +339,19 @@ definitions:
       discoveredCount:
         type: integer
         format: int64
-        description: Number of systems that are discovered and not managed.
+        description: Number of systems that are discovered and not managed by your SystemLink Server.
         example: 100
     required: [discoveredCount]
 
   PendingSystemsSummary:
     title: Pending Systems Summary
-    description: Summary information about the systems that are trying to connect to the server but aren't accepted yet.
+    description: Summary information about the systems attempting to connect to the server but aren't accepted yet.
     type: object
     properties:
       pendingCount:
         type: integer
         format: int64
-        description: Number of systems that are currently pending to be added to the master but have not yet been approved by the master.
+        description: Number of systems pending approval to connect to the server.
         example: 5
     required: [pendingCount]
 
@@ -365,7 +365,7 @@ definitions:
         type: string
         example: Test system
       cpuArch:
-        description: System CPU architecture
+        description: System CPU architecture.
         type: string
         example: armv7l
       deviceClass:
@@ -436,9 +436,9 @@ definitions:
     description: > 
       The state of the job.
        * CANCELED: The job was canceled by the user.
-       * FAILED: The job has completed with failures. The job is failed if at least one of its functions being executed has failed.
-       * SUCCEEDED: The job has completed with success.
-       * INQUEUE: The job is in queue waiting to be dispatched to the system.
+       * FAILED: The job completed with failures. The job fails if at least one of the its functions executing fails.
+       * SUCCEEDED: The job completed successfully.
+       * INQUEUE: The job is in queue, waiting to be dispatched to the system.
        * INPROGRESS: The job was dispatched to the system and is still running.
 
   Job:
@@ -542,7 +542,7 @@ definitions:
             items:
               type: boolean
               example: false
-            description: An array of booleans representing succees values for each executed function.
+            description: An array of booleans representing success values for each executed function.
             example:
               - true
               - true
@@ -566,11 +566,11 @@ definitions:
        * APPROVED: The system is approved but it didn't try to connect yet.
        * DISCONNECTED: The system is disconnected.
        * CONNECTED_REFRESH_PENDING: The system is connected and the refresh job is running. A refresh job is sent to the system every time the connection is established in order to gather the system information.
-       * CONNECTED: The system is connected. The refresh job has completed with success.
-       * CONNECTED_REFRESH_FAILED: The system is connected but the refresh job has failed.
+       * CONNECTED: The system is connected. The refresh job completed successfully.
+       * CONNECTED_REFRESH_FAILED: The system is connected but the refresh job failed.
 
   ManagedSystemActivationData:
-    description: Information about the license of a system
+    description: Information about the license of a system.
     type: object
     properties:
       activated:
@@ -613,12 +613,12 @@ definitions:
         description: The ID of the system.
         example: VMware_Virtual_Platform--SN-VMware-56_4d_d2_08_ce_89_1b_70-e6_32_fc_21_6c_d0_d9_f9--MAC-00-0C-29-D0-D9-F9
       createdTimestamp:
-        description: The time when the system has been registered.
+        description: The time when the system was registered.
         type: string
         format: iso-date-time
         example: '2019-02-19T11:42:25.078Z'
       lastUpdatedTimestamp:
-        description: Last time when the system information has been updated.
+        description: Last time when the system information updated.
         type: string
         format: iso-date-time
         example: '2019-02-19T11:42:25.078Z'
@@ -631,7 +631,7 @@ definitions:
         type: object
         properties:
           lastUpdatedTimestamp:
-            description: Last time when the system activation information has been updated.
+            description: Last time when the system activation information updated.
             type: string
             format: iso-date-time
             example: '2019-02-19T11:42:25.078Z'
@@ -643,7 +643,7 @@ definitions:
         type: object
         properties:
           lastUpdatedTimestamp:
-            description: Last time when the system connection information has been updated.
+            description: Last time when the system connection information updated.
             type: string
             format: iso-date-time
             example: '2019-02-19T11:42:25.078Z'
@@ -671,7 +671,7 @@ definitions:
             osversion: 6.1.7601
             osrelease: 7        
       packages:
-        description: Software packages that are installed on the system.
+        description: Software packages installed on the system.
         $ref: '#/definitions/ManagedSystemDataProperty'
         example:
           lastUpdatedTimestamp: '2019-02-19T11:42:25.078Z'
@@ -695,7 +695,7 @@ definitions:
               group: Infrastructure
               version: 19.5.0.28-0+d28
       feeds:
-        description: Feeds that are configured on the system.
+        description: Feeds configured on the system.
         $ref: '#/definitions/ManagedSystemDataProperty'
         example:
           lastUpdatedTimestamp: '2019-02-19T11:42:25.078Z'
@@ -733,7 +733,7 @@ definitions:
                 tag: 16781312
                 value: false
       groups:
-        description: Groups information about a system. Specifies on which groups is the system assigned to.
+        description: Specifies the groups to which the system is assigned.
         $ref: '#/definitions/ManagedSystemDataProperty'
         example:
           lastUpdatedTimestamp: '2019-02-19T11:42:25.078Z'
@@ -824,7 +824,7 @@ definitions:
     properties:
       groups:
         type: array
-        description: The groups that got created.
+        description: The groups created.
         items:
           $ref: '#/definitions/GroupInfo'
       error:
@@ -897,28 +897,28 @@ definitions:
     properties:
       systemsPending:
         type: object
-        description: An object of system ID to key key/value pairs that are pending approval or rejection. Keys may be pending when they are not pre-seeded via the automated registration process and attempt to connect to the master.
+        description: An object of system ID to key pairs that are pending approval or rejection. Keys may be pending when they are not pre-seeded via the automated registration process and attempt to connect to the master.
         additionalProperties:
           type: string
         example:
           VMware_Virtual_Platform--SN-VMware-56_4d_d2_08_ce_89_1b_70-e6_32_fc_21_6c_d0_d9_f9--MAC-00-0C-29-D0-D9-F6: "-----BEGIN PUBLIC KEY-----↵MIIBIjANBgkqhkiG9w0BAQEFAAOCAQ8AMIIBCgKCAQEAlFgEKiRwWBD/QSDHCpNQ↵Zml9q224YzIxG6reByDV16BNsQoH0lc2bT9LZ7o6xf6dTpQl7shIMc/abSAcl4/F↵VpKguH1yRmOG2Nhout/9wMP6kYBTlThzpop93rlxIo/I7nDkQDz07cMhNWYDvrut↵PVnjRg70xOCAzQRCHRy4GW3TLvfp2WAbCuoQ5oPHQLx+YBSRYDcgfo9jiW8vNzt4↵Bcg72c2PmHEJjGslzJeqcKdRjK0z4/VFgLYfMUBz4LCoy7DohU/uc9ZbHSlOy/BK↵1/S8O1vUyk8DEq+/kQumJWKh1Cl0SWcZGSDkdhQhoTiWqXGb1BBtU7vkxYgQoYjM↵rwIDAQAB↵-----END PUBLIC KEY-----"
       systemsDenied:
         type: object
-        description: An object of system ID to key key/value pairs that have been denied. A key is denied when a system is trying to connect the master with a different key than the one that is accepted. This can occur if two systems have the same ID or if a key is removed from a system.
+        description: An object of system ID to key pairs that have been denied. A key is denied when a system is trying to connect the master with a different key than the one that is accepted. This can occur if two systems have the same ID or if a key is removed from a system.
         additionalProperties:
           type: string
         example:
           VMware_Virtual_Platform--SN-VMware-56_4d_d2_08_ce_89_1b_70-e6_32_fc_21_6c_d0_d9_f9--MAC-00-0C-29-D0-D9-F7: "-----BEGIN PUBLIC KEY-----↵MIIBIjANBgkqhkiG9w0BAQEFAAOCAQ8AMIIBCgKCAQEAlFgEKiRwWBD/QSDHCpNQ↵Zml9q224YzIxG6reByDV16BNsQoH0lc2bT9LZ7o6xf6dTpQl7shIMc/abSAcl4/F↵VpKguH1yRmOG2Nhout/9wMP6kYBTlThzpop93rlxIo/I7nDkQDz07cMhNWYDvrut↵PVnjRg70xOCAzQRCHRy4GW3TLvfp2WAbCuoQ5oPHQLx+YBSRYDcgfo9jiW8vNzt4↵Bcg72c2PmHEJjGslzJeqcKdRjK0z4/VFgLYfMUBz4LCoy7DohU/uc9ZbHSlOy/BK↵1/S8O1vUyk8DEq+/kQumJWKh1Cl0SWcZGSDkdhQhoTiWqXGb1BBtU7vkxYgQoYjM↵rwIDAQAB↵-----END PUBLIC KEY-----"
       systemsApproved:
         type: object
-        description: An object of system ID to key key/value pairs that have been approved. These could have been approved explicitly by the user or pre-seeded via the automated registration process.
+        description: An object of system ID to key pairs that have been approved. These could have been approved explicitly by the user or pre-seeded via the automated registration process.
         additionalProperties:
           type: string
         example:
           VMware_Virtual_Platform--SN-VMware-56_4d_d2_08_ce_89_1b_70-e6_32_fc_21_6c_d0_d9_f9--MAC-00-0C-29-D0-D9-F8: "-----BEGIN PUBLIC KEY-----↵MIIBIjANBgkqhkiG9w0BAQEFAAOCAQ8AMIIBCgKCAQEAlFgEKiRwWBD/QSDHCpNQ↵Zml9q224YzIxG6reByDV16BNsQoH0lc2bT9LZ7o6xf6dTpQl7shIMc/abSAcl4/F↵VpKguH1yRmOG2Nhout/9wMP6kYBTlThzpop93rlxIo/I7nDkQDz07cMhNWYDvrut↵PVnjRg70xOCAzQRCHRy4GW3TLvfp2WAbCuoQ5oPHQLx+YBSRYDcgfo9jiW8vNzt4↵Bcg72c2PmHEJjGslzJeqcKdRjK0z4/VFgLYfMUBz4LCoy7DohU/uc9ZbHSlOy/BK↵1/S8O1vUyk8DEq+/kQumJWKh1Cl0SWcZGSDkdhQhoTiWqXGb1BBtU7vkxYgQoYjM↵rwIDAQAB↵-----END PUBLIC KEY-----"
       systemsRejected:
         type: object
-        description: An object of system ID to key key/value pairs that have been rejected. Keys are rejected explicitly by the user.
+        description: An object of system ID to key pairs that have been rejected. Keys are rejected explicitly by the user.
         additionalProperties:
           type: string
         example:
@@ -971,7 +971,7 @@ definitions:
         required: [keyActions]
 
   SystemPatchRequest:
-    description: System data that has to be assigned to the system.
+    description: System data that needs to be assigned to the system.
     title: System Patch Request
     x-ni-sparse: true
     type: object
@@ -1050,21 +1050,21 @@ definitions:
         description: An array containing information about other packages this package breaks.
         items:
           type: string
-          description: The name of a package that this package breaks. This also contains information about the version of the package.
+          description: The name of a package this package breaks. This also contains information about the version of the package.
           example: "ni-sysapi (>= 18.0.0)"
       conflicts:
         type: array
         description: An array containing information about other packages this package conflicts with.
         items:
           type: string
-          description: The name of a package that this package conflicts with. This also contains information about the version of the package.
+          description: The name of a package this package conflicts with. This also contains information about the version of the package.
           example: "ni-msvcrt-2015 (>= 14.1.4)"
       depends:
         type: array
         description: An array containing information about other packages this package depends on.
         items:
           type: string
-          description: The name of a package that this package depends on. This also contains information about the version of the package.
+          description: The name of a package this package depends on. This also contains information about the version of the package.
           example: "ni-metauninstaller (>= 18.0.1)"
       description:
         type: string
@@ -1083,7 +1083,7 @@ definitions:
         description: An array containing information about other packages this package enhances.
         items:
           type: string
-          description: The name of a package that this package enhances. This also contains information about the version of the package.
+          description: The name of a package this package enhances. This also contains information about the version of the package.
           example: "ni-msiproperties (>= 18.0.1)"
       essential:
         type: boolean
@@ -1121,13 +1121,13 @@ definitions:
       plugin:
         type: integer
         format: int32
-        description: Agent that will be used to install the Package.
+        description: Agent that will be used to install the package.
         example: 2
       predepends:
         type: array
         description: An array containing information about other packages this package predepends.
         items:
-          description: The name of a package that this package predepends. This also contains information about the version of the package.
+          description: The name of a package this package predepends. This also contains information about the version of the package.
           type: string
           example: "ni-teststand-2017-tools (>= 17.0.0)"
       priority:
@@ -1137,17 +1137,17 @@ definitions:
         example: 4        
       provides:
         type: array
-        description: An array containing information about other packages that this package provides.
+        description: An array containing information about other packages this package provides.
         items:
           type: string
           example: "ni-curl (= 17.0.0.49152-0+f0)"
-          description: One of the packages that the containing package provides. This also contains information about the version of the package.      
+          description: One of the packages the containing package provides. This also contains information about the version of the package.      
       recommends:
         type: array
         description: An array containing information about other packages this package recommends.
         items:
           type: string
-          description: The name of a package that this package recommends. This also contains information about the version of the package.
+          description: The name of a package this package recommends. This also contains information about the version of the package.
           example: "ni-systemlink-client-2018.5-realtime-bin (>= 18.5.0)"        
       releaseNotes:
         type: string
@@ -1158,7 +1158,7 @@ definitions:
         description: An array containing information about other packages this package replaces.
         items:
           type: string
-          description: The name of a package that this package replaces. This also contains information about the version of the package.
+          description: The name of a package this package replaces. This also contains information about the version of the package.
           example: "b235b862-6a92-4a04-a8b2-6d71f777de67 (< 16.99.65535)"       
       section:
         type: string
@@ -1175,14 +1175,14 @@ definitions:
         example: "thunar-archive-plugin_0.3.1.bb"
       storeProduct:
         type: boolean
-        description: Whether the package represent a product in the Store or not.
+        description: Whether the package represent a product in the store or not.
         example: true
       suggests:
         type: array
         description: An array containing information about other packages this package suggests.
         items:
           type: string
-          description: The name of a package that this package suggests. This also contains information about the version of the package.
+          description: The name of a package this package suggests. This also contains information about the version of the package.
           example: "ni-systemlink-client-2018.5-realtime-bin (>= 18.5.0)"       
       tags:
         type: string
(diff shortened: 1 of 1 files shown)


systems-management/nisysmgmt.yml
```diff
@@ -0,0 +1,2030 @@
...
+      message:
+        description: The filled in error message.
+        type: string
+        example: The provided query is invalid. 
+      args:
+        description: Positional argument values for the error code.
+        type: array
+        items:
+          type: string
+          example: invalidquery
```
BKnight760: 'invalidquery' should not be in the args array.  The args array is for message inserts.

systems-management/nisysmgmt.yml
```diff
@@ -0,0 +1,2030 @@
...
+      innerErrors:
+        type: array
+        description: Internal errors that occured in the service.
+        items:
+          $ref: '#/definitions/Error'
```
BKnight760, line 1241: For some reason, the example ends up as [ null ] which is not what we want.

systems-management/nisysmgmt.yml
```diff
@@ -0,0 +1,2030 @@
...
+
+  CancelJobRequest:
+    description: Object containing the job ID and a system ID.
+    title: Cancel Job Request
+    type: object
+    properties:
+      jid:
+        description: The job ID of the job to cancel.
+        type: string
+        example: 20190211135234448026_28320
```
BKnight760: 20190211135234448026_28320 gets rendered in scientific notation as "jid": 2.019021113523445e+24 which is probably not what was intended.  This happens in multiple places with this same value.

systems-management/nisysmgmt.yml
```diff
@@ -0,0 +1,2030 @@
...
+          name: Groups
+          description: Array of strings representing group ids.
+          required: true
+          schema:
+            type: array
+            items:
+              type: string
+      responses:
+        200:
+          $ref: '#/responses/Part
```
BKnight760: Should this route be just /v1/query-jobs?  That seems to be the preferred style in other apis.

systems-management/nisysmgmt.yml
```diff
@@ -0,0 +1,2030 @@
...
+    properties:
+      id:
+        type: string
+        description: The ID of the system.
+        example: VMware_Virtual_Platform--SN-VMware-56_4d_d2_08_ce_89_1b_70-e6_32_fc_21_6c_d0_d9_f9--MAC-00-0C-29-D0-D9-F9
+      created_timestamp:
+        description: The time when the system has been registered.
+        type: string
+        format: iso-date-time
+        example: 2019-02-19T11:42:25.078Z
```
BKnight760: The example for all of the timesamp elements in this definition render as an empty object: {} instead of an ISO-8601 string value.

systems-management/nisysmgmt.yml
```diff
@@ -0,0 +1,2030 @@
...
+            items:
+              type: string
+              example: 
+                - testVM
+        required: [last_updated_timestamp, data]
+      properties:
+        description: Properties metadata information about a system.
+        type: object
+        properties:
+          last_updated_timestamp:
```
BKnight760: This 'last_updated_timestamp' needs an example element so that it does not render as "string".  This is needs to be addressed in multiple places in this document.

systems-management/nisysmgmt.yml
```diff
@@ -0,0 +1,2030 @@
...
+          $ref: '#/definitions/Error'
+
+  Unauthorized:
+    description: Not authorized.
+    headers:
+      WWW_Authenticate:
+        description: Information for how to authenticate.
+        type: string
+
+  PartialSuccess:
```
BKnight760, line 55: This partial success response only includes an error object.  What data is included for the items that were successful?  Generally, the partial success response definition needs to be specific to each type of resource / route.

systems-management/nisysmgmt.yml
```diff
@@ -0,0 +1,2030 @@
...
+    type: object
+    properties:
+      hostname:
+        type: string
+        description: Required if ip_address is not provided. A string representing the hostname of the system to register.
+        example: testvm
+      ip_address:
+        type: string
+        description: Required if hostname is not provided. A string representing the IP address of the system to register.
+        example: 1.1.1.1
```
BKnight760: Other places use 10.10.10.10 as the example ip_address.  I personally like that better than 1.1.1.1.

systems-management/nisysmgmt.yml
```diff
@@ -0,0 +1,2030 @@
...
+      responses:
+        200:
+          description: OK
+          schema:
+            $ref: '#/definitions/PendingSystemsSummary'
+        401:
+          $ref: '#/responses/Unauthorized'
+        default:
+          $ref: '#/responses/Error'
+  /v1/get-systems-report:
```
BKnight760: I'd advise against using 'get' in routes names that are HTTP POST verbs instead of HTTP GET to avoid confusion.

systems-management/nisysmgmt.yml
```diff
@@ -0,0 +1,2030 @@
...
+       * Hardware: The report will contain information about the hardware configuration of the system.
+
+  SystemsReportRequest:
+    description: Request containing the report type followed by the list of system IDs.
+    title: Systems Report Request
+    type: object
+    properties:
+      type:
+        $ref: '#/definitions/ReportType'
+      ids:
```
BKnight760: Does this need to be an array of IDs, or could this use the new query syntax as a filter for any matching systems, including a list of IDs?

systems-management/nisysmgmt.yml
```diff
@@ -0,0 +1,2030 @@
...
+          $ref: '#/responses/PartialSuccess'
+        204:
+          description: No Content
+        400:
+          $ref: '#/responses/BadRequest'
+        401:
+          $ref: '#/responses/Unauthorized'
+        default:
+          $ref: '#/responses/Error'
+    get:
```
BKnight760, line 1895: This is odd to have a GET verb for /create-groups.  It seems that there should be a /groups resource on which you would do a POST or GET.

systems-management/nisysmgmt.yml
```diff
@@ -0,0 +1,2030 @@
...
+      description: Remove groups.
+      parameters:
+        - in: body
+          name: Groups
+          description: Array of strings representing group ids.
+          required: true
+          schema:
+            type: array
+            items:
+              type: string
```
BKnight760, line 1938: This needs an example, otherwise it renders as ["string"]

BKnight760 (dismissed):
There are several items that are easily identifiable by opening each route in the swagger editor and viewing the example request and response bodies.  A few routes should be evaluated for naming and usage based on the HTTP verbs that are being used.

systems-management/nisysmgmt.yml
chotea: good catch!

systems-management/nisysmgmt.yml
chotea: I've splitted the initial /groups route into 2 separate routes and I forgot to add the URL for the get route. I can't use a POST to /groups for CreateGroups because we are creating multiple resources, so is RPC instead of REST.

systems-management/nisysmgmt.yml
chotea: This is similar with other Swagger definitions. It seems that it renders the array, the description and inside the array is an empty dict.

systems-management/nisysmgmt.yml
chotea: For almost all routes I think returning the failed items is enough. The only route that could be addressed is /v1/create-groups which could return the generated gids. I will modify it.

systems-management/nisysmgmt.yml
alexweav: The old API had a way of automatically uploading the systems software report to the file ingestion service and returning a file ID instead of the actual report contents. Do we plan to continue supporting this use case in the new API?

systems-management/nisysmgmt.yml
BKnight760: The "Responses" section of the SystemLink HTTP API Guidelines indicates 'partial success' responses should contain more than just the error definition.

https://nitalk.jiveon.com/docs/DOC-410194

systems-management/nisysmgmt.yml
chotea: Only the current AMQP API supports that so we will keep this functionality only in AMQP API.

systems-management/nisysmgmt.yml
chotea: That is right: "If any resources were created, should include a success array of objects in the response". So in this case only the CreateGroups route should contain additional information. The other routes are RPCs so I think they are fine.

systems-management/nisysmgmt.yml
```diff
@@ -0,0 +1,2076 @@
...
+  CancelJobRequest:
+    description: Object containing the job ID and a system ID.
+    title: Cancel Job Request
+    type: object
+    properties:
+      jid:
+        description: The job ID of the job to cancel.
+        type: string
+        example: '20190211135234448026_28320'
+      systemId:
```
skizunov, line 289: You have `request_id` above and `systemId` here. Is there a reason why some use lowercase_with_underscores and others use camelCase?

systems-management/nisysmgmt.yml
```diff
@@ -0,0 +1,2076 @@
...
+        example: VMware_Virtual_Platform--SN-VMware-56_4d_d2_08_ce_89_1b_70-e6_32_fc_21_6c_d0_d9_f9--MAC-00-0C-29-D0-D9-F9
+    required: [jid, systemId]
+
+  ManagedSystemsSummary:
+    title: Managed Systems Summary
+    description: Summary information about managed systems.
+    type: object
+    properties:
+      connected_count:
+        type: integer
```
skizunov, line 301: For Python based codegen (both service and client-side), each time you use `type: integer`, it also needs either `format: int32` or `format: int64`. I would suggest you go through and add these to each and every use of `type: integer` that you have.

If you think about it, it is a silly requirement since in Python 3 everything is just an `int`. But it is a current requirement. I personally tend to lean towards `int64` for most of these, so that any potential C++ codegen makes sense (since in C++, they tend to use size_t for most things, which is 64-bit on 64-bit machines).

systems-management/nisysmgmt.yml
```diff
@@ -0,0 +1,2076 @@
...
+        default: -1
+        maximum: 1000
+        example: 100
+      filter:
+        type: string
+        example: connected.data.state==Connected
+        description: Specifies the filter criteria for resources.
+      projection:
+        type: string
+        example: 'new(id,grains,last_updated_timestamp)'
```
skizunov: Curious, why isn't this an array of strings or enums? Also, in your example, what does the `new` at the front mean (with parenthesis)?

systems-management/nisysmgmt.yml
```diff
@@ -0,0 +1,2076 @@
...
+    properties:
+      skip:
+        description: How many resources to skip.
+        type: integer
+        default: 0
+        example: 100
+      take:
+        description: How many resources to take. Maximum value is 1000.
+        type: integer
+        default: -1
```
skizunov: Instead of defaulting to -1, you can make this optional (default to null) and then have the minimum be 0 (eg by using `minimum: 0`). This way you don't have -1 mean a magic value, whereas the other numbers have a different meaning.

systems-management/nisysmgmt.yml
```diff
@@ -0,0 +1,2076 @@
...
+
+  QueryRequest:
+    description: Request for querying resources. More info here -> https://github.com/StefH/System.Linq.Dynamic.Core/wiki/Dynamic-Expressions and https://stefh.github.io/System.Linq.Dynamic.Core/html/144078bd-d1ad-bece-4daa-7fdd3a7ed7dc.htm
+    title: Query Request
+    properties:
+      skip:
+        description: How many resources to skip.
+        type: integer
+        default: 0
+        example: 100
```
skizunov, line 1315: You could add `minimum: 0` here.

systems-management/nisysmgmt.yml
```diff
@@ -0,0 +1,2076 @@
...
+          description: Feeds to be configured and updated.
+          schema:
+            type: array
+            items:
+              type: string
+              description: The URI of the feed.
+              example: 'http://download.ni.com/support/nipkg/products/ni-package-manager/released'
+      responses:
+        200:
+          $ref: '#/responses/PartialSuccess'
```
skizunov, line 1792: The `PartialSuccess` object does not contain the feeds that were successfully configured nor the feeds that failed.

systems-management/nisysmgmt.yml
```diff
@@ -0,0 +1,2076 @@
+﻿swagger: '2.0'
+info:
+  description: 'Management of distributed systems.'
```
caleyburton: "Systems Management Service HTTP API. Manage and monitor your distributed systems."

systems-management/nisysmgmt.yml
```diff
@@ -0,0 +1,2076 @@
...
+
+  Unauthorized:
+    description: Not authorized.
+    headers:
+      WWW_Authenticate:
+        description: Information for how to authenticate.
+        type: string
+
+  PartialSuccess:
+    description: Partial Success in the case when only a part of the request body metadata information was processed successfully.
```
caleyburton: "Partial Success indicates only part of the request body metadata information was processed successfully."?

systems-management/nisysmgmt.yml
```diff
@@ -0,0 +1,2076 @@
...
+      title: Not Found Request Error Response
+      description: Not found error response containing error information.
+      type: object
+      properties:
+        error:
+          $ref: '#/definitions/Error'
+
+definitions:
+  CreateJobRequest:
+    description: Request describing a job to be sent to systems.
```
caleyburton: "Request describing a job to send to systems."

systems-management/nisysmgmt.yml
```diff
@@ -0,0 +1,2076 @@
...
+        description: An array of strings each representing an execution module function.
+        type: array
+        items:
+          type: string
+          example: pkg.remove
+        example:
+          - pkg.remove
+          - pkg.install
+      arg: 
+        description: An array of arrays of variable type elements that are arguments to the function specified by the "fun" property. If at least one function requires arguments, an array must be provided for each function, even if it is empty. Keyword arguments may be provided as a parameter of type object with at least one property of type boolean, "__kwarg__", that is evaluated to true.
```
caleyburton: "An array of variable type element arrays, which are arguments to the function specified by the "fun" property."?

systems-management/nisysmgmt.yml
```diff
@@ -0,0 +1,2076 @@
...
+              - acl: ''
+        example:
+          - - __kwarg__: true
+              pkgs:
+              - acl: ''
+          - - __kwarg__: true
+              pkgs:
+              - acl-dbg: 1.0.0-0
+      metadata:
+        description: Additional information passed along with the job.
```
caleyburton: "passed to the system with the job" maybe?

systems-management/nisysmgmt.yml
```diff
@@ -0,0 +1,2076 @@
...
+        example:
+          queued: true
+          timeout: 600
+          refresh_minion_cache:
+            grains: true
+    required: [tgt, fun]
+
+  CreateJobResponse:
+    title: Create Job Response
+    description: Response describing a job that was sent to systems.
```
caleyburton: "Response describing a job sent to systems."?

systems-management/nisysmgmt.yml
caleyburton: "... boolean ("_ kwarg _") that evaluates to true."? Sorry, I had to add spaces between the underscores and "kwarg" because markdown removes the underscores and italicize the "kwarg" when they're together.

systems-management/nisysmgmt.yml
```diff
@@ -0,0 +1,2076 @@
...
+          $ref: '#/definitions/SystemCredentials'
+      activate:
+        type: boolean
+        description: Whether to activate the systems or not.
+        default: false
+        example: true
+    required: [systems_credentials]
+
+  SystemCredentials:
+    description: Credentials of a system.
```
caleyburton: "The hostname, IP address, and login information of a system."?

systems-management/nisysmgmt.yml
```diff
@@ -0,0 +1,2076 @@
...
+        description: A string representing the password to use access the system.
+        example: test
+      username:
+        type: string
+        description: A string representing the username to use access the system.
+        example: root
+
+  RegisterSystemsResponse:
+    title: Register Systems Response
+    description: Response of a request to register systems.
```
caleyburton: "Response to a request to register systems to the server." or  "Response to a request to register systems to the Systems Management service."?

systems-management/nisysmgmt.yml
```diff
@@ -0,0 +1,2076 @@
...
+    properties:
+      request_id:
+        description: The ID of the registration request.
+        type: string
+        example: '068530b9-b059-4c0e-9820-9e9cc4e0832c'
+    required: [request_id]
+
+  UnregisterSystemsRequest:
+    title: Unregister Systems Request
+    description: Request to unregister systems.
```
caleyburton: Would it be useful to say "Request to unregister systems from the server" or "Request to unregister system from the Systems Management service"?

systems-management/nisysmgmt.yml
```diff
@@ -0,0 +1,2076 @@
...
+    type: object
+    properties:
+      tgt:
+        description: Array of string representing the IDs of systems to be removed.
```
caleyburton: "Array of strings representing the IDs of systems to remove from SystemLink Server." or "Array of strings representing the IDs of systems to remove from the Systems Management service."?

systems-management/nisysmgmt.yml
```diff
@@ -0,0 +1,2076 @@
...
+        type: array
+        items:
+          type: string
+          example: VMware_Virtual_Platform--SN-VMware-56_4d_30_26_4f_c9_5c_05-b7_68_d1_cd_99_6e_22_16--MAC-02-01-02-03-04-08
+      force:
+        description: Specify as true to remove systems from the database immediately, or false to wait until the unregister job returns from systems. If true, unregister job failures will not be cached.
```
caleyburton: "A Boolean which specifies whether to remove systems from the database immediately (True) or wait until the unregister job returns from systems (False). If True, unregister job failures are not cached."?

systems-management/nisysmgmt.yml
chotea: I will make all properties camel cased.

systems-management/nisysmgmt.yml
chotea: The PartialSuccess response will contain the information about the feeds that were not configured(innerErrors). I don't think we should add any additional information because:
"If any resources were created, should include a success array of objects in the response"
We are not managing feeds with this API, we are not assigning IDs to them, so I don't see them as resources in this API. Our high-level use case is to get the available packages from a list of feeds. Ideally this route and the getAvailablePackages should be merged into one single route but we decided to keep them separated(to leverage the current implementations from SMS) and merge them only if we have enough time before the release.

systems-management/nisysmgmt.yml
chotea: For now we don't want to restrict the API to certain keys when using filtering and projection(and I don't think we can do that since we have the grains dictionary that can contain anything). So we let the user to send a string that should be built based on the JobModel/SystemModel properties. We will make sure that the models aren't changed so we will avoid breaking the API. The "new" operator implies creation of a new object(dict). Without that, the query will return a tuple with the values instead of a dict. So the "new" operator is very important when you want to keep the hierarchy of the response and to avoid flattening. This way the client doesn't have to write custom logic to handle the query response for each different query request.

systems-management/nisysmgmt.yml
```diff
@@ -0,0 +1,2093 @@
...
+        description: An array containing information about other packages this package enhances.
+        items:
+          type: string
+          description: The name of a package that this package enhances. This also contains information about the version of the package.
+          example: "ni-msiproperties (>= 18.0.1)"
+      essential:
+        type: integer
+        format: int32
+        description: Whether the package is essential or not. A value greater than 0 means the package is essential.
+        example: 1
```
skizunov: Is there a reason why this isn't a `boolean` instead of an `integer`? Based on the description, it sounds like a bool to me.

chotea: NIPM C API gives us a numeric value. You can check it here: Perforce\AST\PackageManagement\components\nipkg\export\19.0\19.0.0f0\includes\nipkg_client\shared\constants.h 
The current implementation uses an integer and I thought I can leave it this way. But looking into C# API of NIPM I see that they made it a boolean. So after all it seems it should have been a boolean. I will change it.

skizunov (dismissed):

systems-management/nisysmgmt.yml
```diff
@@ -0,0 +1,2076 @@
...
+          example: VMware_Virtual_Platform--SN-VMware-56_4d_30_26_4f_c9_5c_05-b7_68_d1_cd_99_6e_22_16--MAC-02-01-02-03-04-08
+      failed_ids:
+        description: An array of objects defining failed unregister operations.
+        type: array
+        items:
+          type: object
+          title: Failed Unregister Request
+          properties:
+            id:
+              description: The ID of the system on which unregister operation failed.
```
caleyburton, line 274: "unregister the operation"

systems-management/nisysmgmt.yml
```diff
@@ -0,0 +1,2076 @@
...
+              type: string
+              example: VMware_Virtual_Platform--SN-VMware-56_4d_30_26_4f_c9_5c_05-b7_68_d1_cd_99_6e_22_16--MAC-02-01-02-03-04-08
+            error:
+              $ref: '#/definitions/Error'
+
+  CancelJobRequest:
+    description: Object containing the job ID and a system ID.
```
caleyburton: Could we remove "a" ("the job ID and system ID")?

systems-management/nisysmgmt.yml
```diff
@@ -0,0 +1,2076 @@
...
+    title: Cancel Job Request
+    type: object
+    properties:
+      jid:
+        description: The job ID of the job to cancel.
+        type: string
+        example: '20190211135234448026_28320'
+      systemId:
+        description: The system ID on which the job will be canceled.
```
caleyburton: "The system ID on which the job is canceled."?

systems-management/nisysmgmt.yml
```diff
@@ -0,0 +1,2076 @@
...
+    required: [active_count, failed_count, succeeded_count]
+
+  DiscoveredSystemsSummary:
+    title: Discovered Systems Summary
+    description: Summary information about discovered systems.
+    type: object
+    properties:
+      discovered_count:
+        type: integer
+        description: Number of systems that are discovered and not managed.
```
caleyburton: "not managed by your SystemLink Server" or "not managed by the Systems Management service"?

systems-management/nisysmgmt.yml
```diff
@@ -0,0 +1,2076 @@
...
+        example: 100
+    required: [discovered_count]
+
+  PendingSystemsSummary:
+    title: Pending Systems Summary
+    description: Summary information about the systems that are trying to connect to the server but aren't accepted yet.
```
caleyburton: "Summary information about the systems attempting to connect to the server but aren't accepted yet."

systems-management/nisysmgmt.yml
```diff
@@ -0,0 +1,2076 @@
...
+    type: object
+    properties:
+      pending_count:
+        type: integer
+        description: Number of systems that are currently pending to be added to the master but have not yet been approved by the master.
```
caleyburton: "Number of systems pending approval from the master to connect to the master."?

systems-management/nisysmgmt.yml
```diff
@@ -0,0 +1,2076 @@
...
+    title: System Information
+    description: Network published information about a system.
+    type: object
+    properties:
+      comments:
+        description: System description.
+        type: string
+        example: Test system
+      cpu_arch:
+        description: System CPU architecture
```
caleyburton: Could we add a period here?

systems-management/nisysmgmt.yml
```diff
@@ -0,0 +1,2076 @@
...
+      - Canceled
+      - Failed
+      - Succeeded
+      - InQueue
+      - InProgress
+    example: Succeeded
+    description: > 
+      The state of the job.
+       * Canceled: The job was canceled by the user.
+       * Failed: The job has completed with failures. The job is failed if at least one of its functions being executed has failed.
```
caleyburton: "The job completed with failures. The job fails if at least one of the its functions executing fails."?

systems-management/nisysmgmt.yml
```diff
@@ -0,0 +1,2076 @@
...
+       * Succeeded: The job has completed with success.
```
caleyburton: "The job completed successfully."

systems-management/nisysmgmt.yml
```diff
@@ -0,0 +1,2076 @@
...
+       * InQueue: The job is in queue waiting to be dispatched to the system.
```
caleyburton: Could we add a comma between "queue" and "waiting" (i.e. "queue, waiting")?

systems-management/nisysmgmt.yml
```diff
@@ -0,0 +1,2076 @@
...
+            description: An array of integers representing code values for each executed function.
+            example:
+              - 0
+              - 0
+          success:
+            type: array
+            items:
+              type: boolean
+              example: false
+            description: An array of booleans representing succees values for each executed function.
```
caleyburton: "success"

systems-management/nisysmgmt.yml
```diff
@@ -0,0 +1,2076 @@
...
+      - Connected
+      - ConnectedRefreshFailed
+    example: Connected
+    description: >
+      The connection state of a system.
+       * ActivatedWithoutConnection: The system is activated but the connection is not established yet.
+       * Approved: The system is approved but it didn't try to connect yet.
+       * Disconnected: The system is disconnected.
+       * ConnectedRefreshPending: The system is connected and the refresh job is running. A refresh job is sent to the system every time the connection is established in order to gather the system information.
+       * Connected: The system is connected. The refresh job has completed with success.
```
caleyburton: "The refresh job completed successfully."

systems-management/nisysmgmt.yml
```diff
@@ -0,0 +1,2076 @@
...
+       * ConnectedRefreshFailed: The system is connected but the refresh job has failed.
```
caleyburton: "The system is connected but the refresh job failed."

systems-management/nisysmgmt.yml
```diff
@@ -0,0 +1,2076 @@
...
+
+  ManagedSystemActivationData:
+    description: Information about the license of a system
```
caleyburton: Could we add a period at the end?

systems-management/nisysmgmt.yml
```diff
@@ -0,0 +1,2076 @@
...
+    description: Information about a managed system.
+    title: Managed System
+    type: object
+    properties:
+      id:
+        type: string
+        description: The ID of the system.
+        example: VMware_Virtual_Platform--SN-VMware-56_4d_d2_08_ce_89_1b_70-e6_32_fc_21_6c_d0_d9_f9--MAC-00-0C-29-D0-D9-F9
+      created_timestamp:
+        description: The time when the system has been registered.
```
caleyburton: "was registered" instead?

systems-management/nisysmgmt.yml
```diff
@@ -0,0 +1,2076 @@
...
+        type: string
+        format: iso-date-time
+        example: '2019-02-19T11:42:25.078Z'
+      last_updated_timestamp:
+        description: Last time when the system information has been updated.
```
caleyburton: "updated" instead of "has been updated"?

systems-management/nisysmgmt.yml
```diff
@@ -0,0 +1,2076 @@
...
+      alias:
+        description: The alias of the system.
+        type: string
+        example: testVM
+      activation:
+        description: Activation information.
+        type: object
+        properties:
+          last_updated_timestamp:
+            description: Last time when the system activation information has been updated.
```
caleyburton: "updated" instead of "has been updated"?

systems-management/nisysmgmt.yml
```diff
@@ -0,0 +1,2076 @@
...
+            kernelrelease: 6.1.7601
+            kernelversion: 6.1.7601
+            cpuarch: x86
+            os: Windows
+            os_family: Windows
+            mem_total: 1023
+            osversion: 6.1.7601
+            osrelease: 7        
+      packages:
+        description: Software packages that are installed on the system.
```
caleyburton: remove "that are"?

systems-management/nisysmgmt.yml
```diff
@@ -0,0 +1,2076 @@
...
+              description: Package that indicates the current system is running a 64-bit Windows operating system.
+              displayname: System Package for 64-Bit Windows OS
+              displayversion: 19.5.0
+              url: http://www.ni.com
+              packager: National Instruments <support@ni.com>
+              priority: standard
+              group: Infrastructure
+              version: 19.5.0.28-0+d28
+      feeds:
+        description: Feeds that are configured on the system.
```
caleyburton: remove "that are"?

systems-management/nisysmgmt.yml
```diff
@@ -0,0 +1,2076 @@
...
+            - resource_uri: nisyscfg/enp0s25
+              properties:
+              - type: string
+                tag: 16777216
+                value: "//localhost/nisyscfg/enp0s25"
+              - type: bool
+                tag: 16781312
+                value: false
+      groups:
+        description: Groups information about a system. Specifies on which groups is the system assigned to.
```
caleyburton: "Specifies the groups to which the system is assigned."

systems-management/nisysmgmt.yml
```diff
@@ -0,0 +1,2076 @@
...
+    required: [name]
+
+  CreateGroupsPartialSuccessResponse:
+    title: Create Group Response
+    description: Response containing the IDs of the created groups along with the failures.
+    type: object
+    properties:
+      groups:
+        type: array
+        description: The groups that got created.
```
caleyburton: "The groups created."?

systems-management/nisysmgmt.yml
```diff
@@ -0,0 +1,2076 @@
...
+    required: [request, status_code]
+
+  SystemsKeys:
+    description: An object containing pending, denied, approved, and rejected system keys. If there are no entries in a given category, then the key representing that category is omitted. The value for the keys is in itself a dictionary, with the key representing the system ID and the value the public authorization key used.
+    title: Systems Keys
+    type: object
+    properties:
+      systems_pending:
+        type: object
+        description: An object of system ID to key key/value pairs that are pending approval or rejection. Keys may be pending when they are not pre-seeded via the automated registration process and attempt to connect to the master.
```
caleyburton: Is there another word we can use instead of "key" here? I ask this because "to key key/value pairs" might confuse some customers. Would "locate" or "find" work here? This comment applies to all of the following descriptions with "key key/value pairs" in it.

systems-management/nisysmgmt.yml
```diff
@@ -0,0 +1,2076 @@
...
+              $ref: '#/definitions/KeyAction'
+            key:
+              type: string
+              description: A string representing the public authorization key. If this field is specified, verification that the system ID is associated with this auth key will happen before the desired action is executed.
+              example: -----BEGIN PUBLIC KEY-----↵MIIBIjANBgkqhkiG9w0BAQEFAAOCAQ8AMIIBCgKCAQEAlFgEKiRwWBD/QSDHCpNQ↵Zml9q224YzIxG6reByDV16BNsQoH0lc2bT9LZ7o6xf6dTpQl7shIMc/abSAcl4/F↵VpKguH1yRmOG2Nhout/9wMP6kYBTlThzpop93rlxIo/I7nDkQDz07cMhNWYDvrut↵PVnjRg70xOCAzQRCHRy4GW3TLvfp2WAbCuoQ5oPHQLx+YBSRYDcgfo9jiW8vNzt4↵Bcg72c2PmHEJjGslzJeqcKdRjK0z4/VFgLYfMUBz4LCoy7DohU/uc9ZbHSlOy/BK↵1/S8O1vUyk8DEq+/kQumJWKh1Cl0SWcZGSDkdhQhoTiWqXGb1BBtU7vkxYgQoYjM↵rwIDAQAB↵-----END PUBLIC KEY-----
+          required: [id, action]
+        required: [key_actions]
+
+  SystemPatchRequest:
+    description: System data that has to be assigned to the system.
```
caleyburton: could we say "needs" instead of "has" here?

systems-management/nisysmgmt.yml
```diff
@@ -0,0 +1,2076 @@
...
+      arch:
+        type: string
+        description: The architecture of the package.
+        example: windows_x64
+      breaks:
+        type: array
+        description: An array containing information about other packages this package breaks.
+        items:
+          type: string
+          description: The name of a package that this package breaks. This also contains information about the version of the package.
```
caleyburton: could we remove "that" (i.e. "that this package breaks" --> "package this package breaks")? This removal request applies to the following descriptions that have "that this" in them.

systems-management/nisysmgmt.yml
```diff
@@ -0,0 +1,2076 @@
...
+        type: string
+        description: The maintainer of this package.
+        example: "National Instruments <support@ni.com>"
+      name:
+        type: string
+        description: The name of the package.
+        example: "ni-securityupdate-kb67l8lcqw-killbits"
+      plugin:
+        type: integer
+        description: Agent that will be used to install the Package.
```
caleyburton: could we lowercase "package"?

systems-management/nisysmgmt.yml
```diff
@@ -0,0 +1,2076 @@
...
+        x-ni-data-type: I64
+        description: The size of the package in bytes.
+        example: 750070
+      source:
+        type: string
+        description: The source of this package, if any.
+        example: "thunar-archive-plugin_0.3.1.bb"
+      storeproduct:
+        type: boolean
+        description: Whether the package represent a product in the Store or not.
```
caleyburton: Could we lowercase "store"? Furthermore, should we specific "the NI store"?

systems-management/nisysmgmt.yml
```diff
@@ -0,0 +1,2076 @@
...
+
+  V1Operations:
+    description: V1 operations
+    type: object
+    properties:
+      operations:
+        description: >-
+          Available operations in the v1 version of the API:
+
+          - getJobs: The ability to retrieve the jobs
```
caleyburton: Could we add periods to all of the descriptions here?

systems-management/nisysmgmt.yml
```diff
@@ -0,0 +1,2076 @@
...
+          
+          - createOrCancelJobs: The ability to create or cancel jobs
+          
+          - activateSystems: The ability to activate systems
+          
+          - updateSystemMetadata: The ability to update systems metadata
+          
+          - manageFeeds: The ability to configure feeds and retrieve the available packages
+          
+          - createOrUpdateGroups: The ability to create, update or delete groups
```
caleyburton: "The ability to create, update, or delete groups."

systems-management/nisysmgmt.yml
```diff
@@ -0,0 +1,2076 @@
...
+        description: >-
+          Available operations in the v1 version of the API:
+
+          - getJobs: The ability to retrieve the jobs
+          
+          - getSystems: The ability to retrieve the systems information
+
+          - getGroups: The ability to retrieve the groups
+    
+          - manageSystems: The ability to register/unregister systems
```
caleyburton: Let's say "register or unregister" here?

systems-management/nisysmgmt.yml
```diff
@@ -0,0 +1,2076 @@
...
+    description: Request for querying resources. More info here -> https://github.com/StefH/System.Linq.Dynamic.Core/wiki/Dynamic-Expressions and https://stefh.github.io/System.Linq.Dynamic.Core/html/144078bd-d1ad-bece-4daa-7fdd3a7ed7dc.htm
+    title: Query Request
+    properties:
+      skip:
+        description: How many resources to skip.
+        type: integer
+        default: 0
+        example: 100
+      take:
+        description: How many resources to take. Maximum value is 1000.
```
caleyburton: "The number of resources to return. The maximum value is 1000."

systems-management/nisysmgmt.yml
```diff
@@ -0,0 +1,2076 @@
...
+            $ref: '#/definitions/Operation'
+          createOrUpdateGroups:
+            $ref: '#/definitions/Operation' 
+
+  QueryRequest:
+    description: Request for querying resources. More info here -> https://github.com/StefH/System.Linq.Dynamic.Core/wiki/Dynamic-Expressions and https://stefh.github.io/System.Linq.Dynamic.Core/html/144078bd-d1ad-bece-4daa-7fdd3a7ed7dc.htm
+    title: Query Request
+    properties:
+      skip:
+        description: How many resources to skip.
```
caleyburton: "The number of resources to skip."

systems-management/nisysmgmt.yml
```diff
@@ -0,0 +1,2076 @@
...
+      take:
+        description: How many resources to take. Maximum value is 1000.
+        type: integer
+        default: -1
+        maximum: 1000
+        example: 100
+      filter:
+        type: string
+        example: connected.data.state==Connected
+        description: Specifies the filter criteria for resources.
```
caleyburton: "The criteria used to filter resources."?

systems-management/nisysmgmt.yml
```diff
@@ -0,0 +1,2076 @@
...
+      projection:
+        type: string
+        example: 'new(id,grains,last_updated_timestamp)'
+        description: Specifies the projection for resources.
+      orderBy:
+        type: string
+        example: 'last_updated_timestamp descending'
+        description: Specifies the ordering of returned data.
```
caleyburton: "The order in which data returns."?

systems-management/nisysmgmt.yml
```diff
@@ -0,0 +1,2076 @@
...
+        maximum: 1000
+        example: 100
+      filter:
+        type: string
+        example: connected.data.state==Connected
+        description: Specifies the filter criteria for resources.
+      projection:
+        type: string
+        example: 'new(id,grains,last_updated_timestamp)'
+        description: Specifies the projection for resources.
```
caleyburton: What does "projection" mean here? Are we creating the resource request with this parameter?

systems-management/nisysmgmt.yml
```diff
@@ -0,0 +1,2076 @@
...
+      x-ni-privilege: Write
+      x-ni-operation: createOrCancelJobs
+      tags: [jobs]
+      summary: Create a job
+      description: Create a job.
+      parameters:
+        - in: body
+          name: CreateJobRequest
+          required: true
+          description: The definition of job to be executed
```
caleyburton: Could we add a period to then end?

systems-management/nisysmgmt.yml
```diff
@@ -0,0 +1,2076 @@
...
+        default:
+          $ref: '#/responses/Error'
+  /v1/cancel-jobs:
+    post:
+      operationId: CancelJobs
+      x-ni-privilege: Write
+      x-ni-operation: createOrCancelJobs
+      tags: [jobs]
+      summary: Cancel jobs
+      description: Cancel jobs that are running on systems. This won't stop the jobs execution on systems if the jobs have already been dispatched to the systems.
```
caleyburton: "This won't stop the jobs from executing on systems"

systems-management/nisysmgmt.yml
```diff
@@ -0,0 +1,2076 @@
...
+        default:
+          $ref: '#/responses/Error'
+  /v1/get-jobs-summary:
+    get:
+      operationId: GetJobsSummary
+      x-ni-privilege: Read
+      x-ni-operation: getJobs
+      tags: [jobs]
+      summary: Get jobs summary
+      description: Get the number of active, failed and succeeded jobs.
```
caleyburton: "active, failed, and succeeded jobs"

systems-management/nisysmgmt.yml
```diff
@@ -0,0 +1,2076 @@
...
+      summary: Get groups
+      description: Get groups information.
+      parameters:
+        - in: query
+          name: gid
+          type: string
+          description: A group ID.
+      responses:
+        200:
+          descript
```
caleyburton, line 2018: Missing a period.

caleyburton (dismissed):

chotea:
> There are several items that are easily identifiable by opening each route in the swagger editor and viewing the example request and response bodies. A few routes should be evaluated for naming and usage based on the HTTP verbs that are being used.

skizunov (dismissed):

systems-management/nisysmgmt.yml
```diff
@@ -0,0 +1,2092 @@
...
+        description: The number of resources to return. The maximum value is 1000.
+        type: integer
+        format: int64
+        minimum: 0
+        maximum: 1000
+        example: 100
+      filter:
+        type: string
+        example: connected.data.state==CONNECTED
+        description: The criteria used to filter resources.
```
vdeac: For the filter, I recommend that our public APIs are explicit and describe both the syntax of the language and explicitly state the properties which can be used in the filter.

Take a look at the [Query-Assets](http://demo.systemlink.io/niapis/?urls.primaryName=Asset%20Performance%20Management#/assets/query-assets ) request and you can use the same text here too.

systems-management/nisysmgmt.yml
```diff
@@ -0,0 +1,2092 @@
...
+        default:
+            $ref: '#/responses/Error'
+  /v1/jobs:
+    post:
+      operationId: CreateJob
+      x-ni-privilege: Write
+      x-ni-operation: createOrCancelJobs
+      tags: [jobs]
+      summary: Create a job
+      description: Create a job.
```
vdeac: Would it make sense to also mention what is the effect of the route (creates a job and returns the newly created job)?

systems-management/nisysmgmt.yml
```diff
@@ -0,0 +1,2092 @@
...
+      description: Request to unregister systems.
+      parameters:
+        - in: body
+          name: UnregisterSystemsRequest
+          description: Request data to unregister systems.
+          required: true
+          schema:
+            $ref: '#/definitions/UnregisterSystemsRequest'
+      responses:
+        200:
```
vdeac: Seems this could also return partial success, if only a subset of systems were unregistered.

systems-management/nisysmgmt.yml
```diff
@@ -0,0 +1,2092 @@
...
+          $ref: '#/responses/Error'
+  /v1/get-discovered-systems:
+    get:
+      operationId: GetDiscoveredSystems
+      x-ni-privilege: Read
+      x-ni-operation: getSystems
+      tags: [systems]
+      summary: Get discovered systems
+      description: Get network discovered systems.
+      responses:
```
vdeac, line 1726: Doesn't this route need a 401 and default responses?

systems-management/nisysmgmt.yml
```diff
@@ -0,0 +1,2092 @@
...
+          description: No Content.
+        400:
+          $ref: '#/responses/BadRequest'
+        401:
+          $ref: '#/responses/Unauthorized'
+        default:
+          $ref: '#/responses/Error'
+  /v1/groups:
+    get:
+      oper
```
vdeac, line 1907: It seems this would need a 404 Not Found response if no group matches the provided id

vdeac, line 2024: 404 Not found ?

vdeac, line 2133: 404 not found if no packages match the request parameters?

vdeac (dismissed):
Overall looks good.
I would advise to add more documentation around the filtering/projection parameters in the requests and also add missing responses to some of the routes.

systems-management/nisysmgmt.yml
chotea: I don't feel that we should return a 404 in this case. The input is always an array of credentials, so having one element in that array is a particular case. Checking if the array contains only one element to return 404 in error case would complicate the handling of the response on the client side since the response of a request containing more than 1 element would come with 200 and a specific schema.

BKnight760 (approved):

PR #51, opened 2019-04-26
API Review - Test Monitor Advanced Query Language

alexweav:

Specifies the Test Monitor API changes to be made for the new Advanced Query Filter feature.

It adds the new routes for F130029

Verified that the generated documentation looks correct.
Verified that the generated C# clients correctly implement the new routes.

### DIFF
--- a/test-monitor/nitestmonitor.yml
+++ b/test-monitor/nitestmonitor.yml
@@ -1174,7 +1174,9 @@ paths:
     post:
       tags: [results]
       summary: Advanced Test Result Query
-      description: Uses the Dynamic Linq query language to specify a filter and return test results. An empty request body queries all test results.
+      description: >-
+        Uses the Dynamic Linq query language to specify a filter and return
+        test results. An empty request body queries all test results.
       operationId: query-results-v2
       x-ni-operation: queryResults
       x-ni-privilege: Read
@@ -1383,7 +1385,9 @@ paths:
     post:
       tags: [steps]
       summary: Advanced Test Step Query
-      description: Uses the Dynamic Linq query language to specify a filter and return test steps. An empty request body queries all test steps.
+      description: >-
+        Uses the Dynamic Linq query language to specify a filter and return
+        test steps. An empty request body queries all test steps.
       operationId: query-steps-v2
       x-ni-operation: querySteps
       x-ni-privilege: Read
--- a/test-monitor/nitestmonitor.yml
+++ b/test-monitor/nitestmonitor.yml
@@ -1216,7 +1216,7 @@ paths:
           $ref: '#/responses/Unauthorized'
         default:
           $ref: '#/responses/Error'
-  /v1/results:
+  /v2/results:
     get:
       tags: [results]
       summary: Gets test results
@@ -1224,6 +1224,7 @@ paths:
       operationId: get-results
       x-ni-operation: getResults
       x-ni-privilege: Read
+      x-ni-oldest-compatible-version: 1
       parameters:
         - in: query
           name: skip
@@ -1251,6 +1252,7 @@ paths:
       operationId: update-results
       x-ni-operation: updateResults
       x-ni-privilege: Write
+      x-ni-oldest-compatible-version: 1
       parameters:
         - in: body
           name: requestBody
@@ -1290,6 +1292,7 @@ paths:
       operationId: create-results
       x-ni-operation: createResults
       x-ni-privilege: Write
+      x-ni-oldest-compatible-version: 1
       parameters:
         - in: body
           name: requestBody
@@ -1316,7 +1319,7 @@ paths:
           $ref: '#/responses/Unauthorized'
         default:
           $ref: '#/responses/Error'
-  /v1/results/{resultId}:
+  /v2/results/{resultId}:
     delete:
       tags: [results]
       summary: Deletes a test result
@@ -1324,6 +1327,7 @@ paths:
       operationId: DeleteResult
       x-ni-operation: deleteResult
       x-ni-privilege: Write
+      x-ni-oldest-compatible-version: 1
       parameters:
         - in: path
           name: resultId
@@ -1342,7 +1346,7 @@ paths:
           $ref: '#/responses/Unauthorized'
         default:
           $ref: '#/responses/Error'
-  /v1/delete-results:
+  /v2/delete-results:
     post:
       tags: [results]
       summary: Deletes test results
@@ -1350,6 +1354,7 @@ paths:
       operationId: delete-results
       x-ni-operation: deleteResults
       x-ni-privilege: Write
+      x-ni-oldest-compatible-version: 1
       parameters:
         - in: body
           name: requestBody
@@ -1427,7 +1432,7 @@ paths:
           $ref: '#/responses/Unauthorized'
         default:
           $ref: '#/responses/Error'
-  /v1/steps:
+  /v2/steps:
     get:
       tags: [steps]
       summary: Gets test steps
@@ -1435,6 +1440,7 @@ paths:
       operationId: get-steps
       x-ni-operation: getSteps
       x-ni-privilege: Read
+      x-ni-oldest-compatible-version: 1
       parameters:
         - in: query
           name: skip
@@ -1462,6 +1468,7 @@ paths:
       operationId: update-steps
       x-ni-operation: updateSteps
       x-ni-privilege: Write
+      x-ni-oldest-compatible-version: 1
       parameters:
         - in: body
           name: requestBody
@@ -1488,6 +1495,7 @@ paths:
       operationId: create-steps
       x-ni-operation: createSteps
       x-ni-privilege: Write
+      x-ni-oldest-compatible-version: 1
       parameters:
         - in: body
           name: requestBody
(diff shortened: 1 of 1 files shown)


test-monitor/nitestmonitor.yml
```diff
@@ -1184,6 +1204,40 @@ paths:
           $ref: '#/responses/Unauthorized'
         default:
           $ref: '#/responses/Error'
+  /v2/query-results:
+    post:
+      tags: [results]
```
skizunov, line 1175: Would you change the `tags` in the `/v1/query-results` path to be:
```
tags: [deprecated]
```
For Python clients, this will move the v1 API into:
systemlink/clients/nitestmonitor/api/deprecated_api.py (DeprecatedApi)
So that it is not something that is likely to be used by a new user of the Python client.

test-monitor/nitestmonitor.yml
```diff
@@ -1383,6 +1437,40 @@ paths:
           $ref: '#/responses/Unauthorized'
         default:
           $ref: '#/responses/Error'
+  /v2/query-steps:
+    post:
+      tags: [steps]
```
skizunov, line 1391: Similarly, would you change the tags in the `/v1/query-steps` tags to be:
```
tags: [deprecated]
```

test-monitor/nitestmonitor.yml
```diff
@@ -1184,6 +1204,40 @@ paths:
...
+          type: integer
+          format: int32
+          default: 0
+          minimum: 0
+        - in: query
+          name: take
+          type: integer
+          format: int32
+          default: -1
+          minimum: -1
```
skizunov: Both skip and take need a description. I would also suggest changing take to be have valid values of "null" and then integers >=0, much like this API (which was based on a suggestion from Paul):
https://github.com/ni/systemlink-OpenAPI-documents/blob/master/user-data/niuserdata.yml#L792

test-monitor/nitestmonitor.yml
```diff
@@ -1383,6 +1437,40 @@ paths:
...
```
skizunov: Same: Both skip and take need a description. I would also suggest changing take to be have valid values of "null" and then integers >=0, much like this API (which was based on a suggestion from Paul):
https://github.com/ni/systemlink-OpenAPI-documents/blob/master/user-data/niuserdata.yml#L792

skizunov (changes requested):

test-monitor/nitestmonitor.yml
```diff
@@ -632,6 +632,26 @@ definitions:
...
+    title: Advanced Query Object for Test Data
+    type: object
+    properties:
+      filter:
+        description: The query filter, in Dynamic Linq
+        type: string
+        example: (operator == "user1") || ((programName != "MyProgram") && (totalTimeInSeconds < @0))
+      substitutions:
+        description: >-
+          Substitutions to make into the query filter expression. Substitutions in the query expression
```
caleyburton: "Makes substitutions in the query filter expression."

caleyburton: "Substitutions for the query expression are indicated by non-negative integers that are prefixed with the "at" symbol."

caleyburton (changes requested):

test-monitor/nitestmonitor.yml
```diff
@@ -1187,16 +1187,26 @@ paths:
             $ref: '#/definitions/AdvancedQuery'
         - in: query
           name: skip
+          description: >-
+            The number of results to skip when paging.
```
caleyburton: Potential descriptions to replace this one with:
- The number of results to skip when returning results.
- The number of results to omit from the returned results.
- The number of results to not include in the returned results.

caleyburton (changes requested):

test-monitor/nitestmonitor.yml
alexweav: What are your thoughts on `The number of matching results to skip when querying.`?

test-monitor/nitestmonitor.yml
```diff
@@ -1150,33 +1170,43 @@ paths:
           description: Not Found
           schema:
             $ref: '#/definitions/Error'
-  /v1/query-results:
+  /v2/query-results:
```
spanglerco, line 1173: Our current guidelines mean you also need to make v2 of all of the other routes as well. Either by copying them all (eww) or making the version a parameter like for the /{version} route.

spanglerco (changes requested):

test-monitor/nitestmonitor.yml
caleyburton: I like that! 👍

test-monitor/nitestmonitor.yml
```diff
@@ -632,6 +632,26 @@ definitions:
         type: array
         items:
           $ref: '#/definitions/ResultSortDefinitionObject'
+  AdvancedQuery:
+    title: Advanced Query Object for Test Data
+    type: object
+    properties:
+      filter:
+        description: The query filter, in Dynamic Linq
```
vdeac, line 640: What we did, for the asset filtering APIs, is enumerate all properties that can be used in the filter. I think this is important as we need to document all possible/supported properties which can participate in filtering. 
You can also take the introductory query text we have written for the query filter. take a look at the model of this route: http://demo.systemlink.io/niapis/?urls.primaryName=Asset%20Performance%20Management#/assets/query-assets

This way all our APIs could be aligned with regards to the documentation we provide for our filters.

test-monitor/nitestmonitor.yml
vdeac: Copying all seems a bit extreme. An improvement we can make (in the future, product-wide) to our API versioning is to set a parameter in the header of the request, thus leaving the route as-is and clients would explicitly request a specific version of the route. 
Thoughts?

spanglerco: We're currently investigating this process:

- For routes that break compatibility, copy the current version of the route and give it the deprecated tag, then make a vnext version of the route with the change.
- All other routes that are still compatible, bump to vnext and add an attribute indicating it is still compatible with a previous range of versions. For example, something like `x-ni-oldest-compatible-version: 1`.
- Bump the service version at the top in the info section to vnext.

Then we can use the compatibility attribute during codegen to automatically duplicate the routes in the INI file without needing to duplicate them in the documentation or the code.

caleyburton (dismissed):

test-monitor/nitestmonitor.yml
alexweav: I would like to avoid having to duplicate generated code (or docs) for routes which maintained backwards compatibility across the version change. There's an item on our backlog to make the necessary codegen changes. I'll keep this review alive until that is complete.

test-monitor/nitestmonitor.yml
```diff
@@ -1150,33 +1170,43 @@ paths:
...
             $ref: '#/definitions/Error'
-  /v1/query-results:
+  /v2/query-results:
     post:
       tags: [results]
-      summary: Queries test results
-      description: Uses input criteria to filter and return test results. An empty request body queries all test results.
-      operationId: query-results
+      summary: Advanced Test Result Query
+      description: Uses the Dynamic Linq query language to specify a filter and return test results. An empty request body queries all test results.
```
BKnight760: Can we use a multi-line description here?   ( ">-" ) The text is wide and does not wrap.

test-monitor/nitestmonitor.yml
```diff
@@ -1349,33 +1379,43 @@ paths:
...
           $ref: '#/responses/Error'
-  /v1/query-steps:
+  /v2/query-steps:
     post:
       tags: [steps]
-      summary: Queries test steps
-      description: Uses input criteria to filter and return test steps. An empty request body queries all test steps.
-      operationId: query-steps
+      summary: Advanced Test Step Query
+      description: Uses the Dynamic Linq query language to specify a filter and return test steps. An empty request body queries all test steps.
```
BKnight760: Multi-line, please.

BKnight760 (dismissed):

skizunov (dismissed):

test-monitor/nitestmonitor.yml
alexweav: After some discussion, our plan is to create documentation for the query language in Markdown, which will live in this repo next to the swagger doc. The documentation in the swagger doc itself will contain a simple example, plus a link to the markdown document here for further info. The contents of the markdown document will be adapted from the query text linked above, as well as the existing documentation for the query language.

We have a separate work item to track this documentation, so it won't be included as part of this review.

test-monitor/nitestmonitor.yml
alexweav: The codegen changes are up for review now.

spanglerco (commented):
Consider updating the example response for the GET / route to show at least a v2 (probably don't need to show both v1 and v2?)

spanglerco (approved):

skizunov (approved):

maxgax (approved):

BKnight760 (approved):

alexweav:
@spanglerco I'll address your above comment in a future review. Added a task to track it.

PR #53, opened 2019-05-21
Modified tag historian API for discussion

pvallone:
Just some quick modifications removing some of the routes we discussed, and adding continuation tokens to the query route.

I don't plan on submitting this; just putting this up here for discussion and information sharing.

I know SLC has a route for inserting values, but the historian works behind the scenes on SLS, so I wouldn't want that to be a public route which we document in swagger.

- Removing routes which reference individual values by ID
- Removing the ability to annotate values with properties
- Removing skip/totalCount in favor of continuation tokens


gergo-papp:
These are just some ideas/comments:
- We could make the insert route a cloud-specific one. And we could even add possible limitations in the swagger document too.
- In the future we could add explanation about decimation and document how we actually do it (random, first, min-max... or some other weird smart decimation). Or even add an option for selecting teh kind of decimation (e.g. choose between fast or accurate decimation)

pvallone:
> * We could make the insert route a cloud-specific one. And we could even add possible limitations in the swagger document too.

We could make it cloud only. Do you folks expect users to call that route? I would have thought you would expect folks to push updates to the tag service's existing routes for updating tag values, and then have your service automatically call the insert-values route on behalf of your users if the modified tag(s) have retention enabled. That's sort of how the existing SLS historian works.

> * In the future we could add explanation about decimation and document how we actually do it (random, first, min-max... or some other weird smart decimation). Or even add an option for selecting teh kind of decimation (e.g. choose between fast or accurate decimation)

Right, in my email I had suggested we provide a "decimationStrategy" input on the query route. The options could be "RANDOM_SAMPLING" and "FULL_SCAN". "FULL_SCAN" would be the "smart" approach we'd discussed, where we return the entry/exit/min/max for each chunk of data that we have in the window. I had suggested we document that the service will consider the requested decimation strategy a suggestion, and fall back to random sampling if there's too much data to decimate with full scan decimation in the user's requested window. "RANDOM_SAMPLING" could be just plain old random sampling as you folks are doing now, or it could sample the buckets and return entry/exit/min/max from each sampled bucket. This would make it a little more accurate at no meaningful expense to performance (especially if entry/exit/min/max is precalculated on the bucket level). Michael and Fred had thought that this sounded like a good idea. I'd be curious to get @tschmittni's take on this.

tag-historian/nitaghistorian.yml
```diff
@@ -316,12 +256,6 @@ definitions:
               properties:
                 note: This value is an outlier and needs investigation.
               expirationDate: '2020-03-13T09:51:38Z'
```
tschmittni: We should probably remove all metadata fields: 
- expirationDate
- type

tschmittni:
_We could make it cloud only. Do you folks expect users to call that route?_

No, this route is not supposed to be called by users. We certainly could provide it because it is protected just as any other route. But this basically makes changes/optimizations a lot harder. And I don't really see a use-case for this route. Customers are supposed to call update-tag-value on the tag-service instead.

_about decimationStrategy_

I really like this feature, but I don't think we should add any decimation strategy which we cannot support in a performant way. This will just cause a lot of problems. We should just add APIs which can be implemented to respond in the ms range and which scale with the number of data points.

I outlined here https://github.com/ni-kismet/helium-taghistorianservices/pull/171 a couple of options to support min/max/avg/etc... (FULL_SCAN) decimation strategies in a way which allows us to implement it in a performant and extensible way. 

My suggestion is to perform decimated history queries based on pre-computed values instead of querying across all the data and compute on-the-fly or calculating values during insert. We just run a background job to calculate the values for our supported decimation strategies and store them so we can highly optimize our read queries.

This has the following advantages:
- Insert performance is not affected and we can still ingest huge amounts of data
- Decimated queries are highly optimized and can return data within milliseconds 
- We can support different strategies (and not limit ourselves to the decimations which our database supports)

Tradeoffs:
- Background job needs to be implemented (which we anyhow already have for deleting old values)
- We either have to disallow inserting values in the past or we need to keep track which documents have been modified

_FULL_SCAN_

Maybe we can come up with a different name for the FULL_SCAN decimation strategy? :)

pvallone:
> No, this route is not supposed to be called by users. 

Then it's probably best to keep it out of the swagger document, yeah. If we don't document it, we can break it.

> We can support different strategies (and not limit ourselves to the decimations which our database supports)

That's a big win over doing all of the decimation in the schema at insert time as I had suggested.

> Decimated queries are highly optimized and can return data within milliseconds

I like the idea of doing decimated queries against precomputed values so that all reads can complete in milliseconds. When I was prototyping on SLS, the fact that the decimated query API lets you specify an arbitrary decimation "number" made it hard to efficiently do queries, even against precomputed values. What were you planning on precomputing to efficiently handle decimated queries? We can precompute entry/exit/min/max at the minute, hour, day, week, month, year, etc. level, but it still might take a while to break that into the appropriate number of chunks based on the specified decimation value that the user provided. I guess we can wait and see what decimated queries against precomputed values look like in SLC with the background thread approach, and how they perform, but based on what I was seeing when I was prototyping on SLS I think we're still going to have multi-second decimated query times over  windows saturated with lots of data even if we precompute aggregates.

If we instead changed the "decimation" query parameter from an integer to an enum with values like MINUTES, HOURS, DAYS, WEEKS, MONTHS, then we could handle queries much more efficiently. InsightCM's tag historian used to do this (and they might still). That would be a more substantial API change though.

> We either have to disallow inserting values in the past or we need to keep track which documents have been modified

I think we need to support inserting old values, as Mark is going to want folks to be able to migrate from old NI products like Citadel to the tag historian. We would also need to insert old values when migrating folks from the SLS historian to the SLC historian.

> Maybe we can come up with a different name for the FULL_SCAN decimation strategy? :)

Gergo had been calling it "accurate" decimation, and I kind of like that.

tschmittni:
_If we instead changed the "decimation" query parameter from an integer to an enum with values like MINUTES, HOURS, DAYS, WEEKS, MONTHS, then we could handle queries much more efficiently. _

couldn't we do this implicitly behind the scenes? basically depending on range size between start and end time (or the first and last data point) we could calculate whether we pick our monthly, daily or hourly pre-computed decimated values...

_I think we need to support inserting old values, as Mark is going to want folks to be able to migrate from old NI products like Citadel to the tag historian._

We could provide specific migration APIs for that...

pvallone:
> _If we instead changed the "decimation" query parameter from an integer to an enum with values like MINUTES, HOURS, DAYS, WEEKS, MONTHS, then we could handle queries much more efficiently. _
> 
> couldn't we do this implicitly behind the scenes? basically depending on range size between start and end time (or the first and last data point) we could calculate whether we pick our monthly, daily or hourly pre-computed decimated values...

Yes, but it might need to be smarter than just how much time is within the window. If a user only wrote one value per hour for a year, and specified a decimation value of "1000" and a window size of half a year, then we probably wouldn't want to decimate at the day level, for instance. When I was prototyping on SLS, I used a heuristic based on the approximate number of distinct seconds which we had data stored for within the window. I kept track of the number of distinct seconds which had data on each hour document using a "count" field. To decide how to decimate the data, I would ask Mongo to sum up the "count" field of all of the documents within the window. Based on the value of the "count" sum, I would decide whether or not to decimate at the minute or hour level, with the idea being that if there was too much second-level data in the window, I would fall back to hour-level decimation. We could fall back to day level decimation etc. in that way too.

My main point was that, even if we're precomputing aggregates at the minute, hour, day, week, etc. level, we're still going to have to break the precomputed aggregates for the window into "decimation" number of chunks, and calculate the entry/exit/min/max value for each chunk service-side. I suspect we'll still have situations where queries can take multiple seconds to complete in this manner due to this overhead and the fact that we can't precompute all possible queries for all possible "decimation" values. That's what I was seeing when I prototyped this on SLS, at least. We can mitigate this by caching decimated queries. Maybe it's best to wait and see how things perform in Elixir.

> We could provide specific migration APIs for that...

That would be fine with me. As long as there's a way for customers to get old data into the historian, and for me to migrate SLS historian data in to the SLC historian, then we're good.

tag-historian/nitaghistorian.yml
pvallone: I removed the expirationDate references.

I think "type" could still be useful for clients. Tag viewer shows historical tag values for "string" tags in a table instead of a graph, for example, and it seems less than ideal for clients like that to have to query the tag service for the type info, or detect the type of a value. It isn't possible to change the type of a tag, so it seems OK to attach type info to the values. Any thoughts?

tschmittni:
_We could fall back to day level decimation etc. in that way too._

Yeah, good point. I would say we just select the granularity (e.g. HOUR or DAY) based on the size of the range between first and last data point (within users' start/end range). Then we perform the decimation and if we don't have enough data points, we just fall back to a more granular level.

_calculate the entry/exit/min/max value for each chunk service-side._

If we can efficiently query for the pre-decimated data points, I don't think calculating server-side min/max values is the actual problem. This should be rather fast.

pvallone:
> Then we perform the decimation and if we don't have enough data points, we just fall back to a more granular level.

Sounds reasonable.

> I don't think calculating server-side min/max values is the actual problem. This should be rather fast.

Yeah, now that I think about it, most of my query time was spent querying and projecting the decimated values out of Mongo, not calculating the chunk min/max etc. service-side.

tag-historian/nitaghistorian.yml
```diff
@@ -228,26 +173,40 @@ definitions:
...
-        default: 0
       take:
         description: Limits the returned historical values to the specified number of values. Use
-          with the skip property to implement pagination. Only applies when not specifying a decimation.
+          with the continuationToken property to implement pagination. Only applies when not specifying a decimation.
         type: integer
         minimum: 0
         default: 0
         example: 1
+      continuationToken:
```
tschmittni: I would say a single continuationToken with multiple tag paths does not make a lot of sense...

We should probably change "paths" to only support a single tag "path".

tag-historian/nitaghistorian.yml
```diff
@@ -228,26 +173,40 @@ definitions:
...
+        type: string
+        description: For non-decimated queries, a value generated by the server which should be sent
+          in the next request to retrieve the next page of values.
       decimation:
         description: When non-zero, the returned historical values are decimated across requested time
           span divided into the specified number of equal-sized intervals.
         type: integer
         minimum: 0
         default: 0
+      decimationStrategy:
```
tschmittni: Let's remove this for now and keep it simple. We try to use the min/max strategy and don't support random sampling for now

tag-historian/nitaghistorian.yml
```diff
@@ -228,26 +173,40 @@ definitions:
...
         example: '2018-09-04T18:45:08Z'
         default: '0001-01-01T00:00:00Z'
-      skip:
-        description: Limits the returned historical values to those after the given number of values.
-          Use with the take property to implement pagination. Only applies when not specifying a
-          decimation.
-        type: integer
-        minimum: 0
-        default: 0
       take:
```
tschmittni: We should remove take if we are not sure about whether we need it or not

tag-historian/nitaghistorian.yml
spanglerco: The LabVIEW API currently has a multi-read history VI. We'll need to determine if this means we would remove that VI entirely, simulate it client-side by making a request per tag, or continue supporting it in the HTTP API. @prestwick, what do you think about potentially dropping support for the Multi-Read History?

spanglerco: Also, in the current API, skip and take doesn't apply to decimated reads. Will that still be true with the continuation token? Both the web UI and dashboards will want to be able to read decimated history for multiple tags for display on a single graph.

pvallone: > I would say a single continuationToken with multiple tag paths does not make a lot of sense...

It seems like the service could process the tag paths in order, and assume that whatever tag comes after the one referenced in the current continuation token is the next one to process.

tag-historian/nitaghistorian.yml
tschmittni: It is just a weird mix between historical values and metadata. It would be better to keep responsibilities between tag service and tag historian clear. It also leads to lot of redundant type information for every single value. I would expect clients to know what data a certain tag produces. For example, the tag viewer already retrieves the tag metadata first

tag-historian/nitaghistorian.yml
tschmittni: Any process on this one? We can encode the necessary information for each requested tag path in the continuation token. It just seems quite weird and overly complex to iterate over multiple tag paths.

_It seems like the service could process the tag paths in order, and assume that whatever tag comes after the one referenced in the current continuation token is the next one to process._

Yeah, but isn't the API supposed to return multiple results at once, meaning if you ask for tag "A" and "B", the service returns the first 100 values of "A" and first 100 values of "B" in the first response?

tag-historian/nitaghistorian.yml
pvallone: Paul and I's thinking is that the query-history response isn't useful without the type information. The values will look like strings, and might actually just be strings. Requiring clients to be stateful and keep a map of tag paths to types to be able to interpret the values seems inconvenient. Carson tells me that dashboard builder would eventually like to use the type information in the response as well.

tag-historian/nitaghistorian.yml
pvallone: We're most concerned about being able to supply multiple tag paths for decimated queries, as users could have dashboards with tiles containing many tags. Chrome and Firefox limit users to 6 concurrent requests per host, so being able to specify multiple tags per request is useful in that context. The service could process each tag in the query in parallel.

We're less sure about undecimated queries. One option could be to change the behavior for undecimated queries such that the service queries the tags one at a time, rather than all at once, as we agree that the current behavior is strange. Continuation tokens would be more straightforward with that model. Another option would be to split the undecimated query stuff out into a separate route, and possibly make that route only support querying for one tag at a time.

> Also, in the current API, skip and take doesn't apply to decimated reads. Will that still be true with the continuation token?

To reiterate Paul's question above: Will continuation tokens apply to decimated queries? It seems like they shouldn't.

tschmittni: _Will continuation tokens apply to decimated queries?_
They would not apply to decimated queries.

_The service could process each tag in the query in parallel._
The client can just send the requests in parallel. And we can basically start displaying tiles when the first request gets returned and not when the slowest query finally finished.
This also makes it easier to apply throttling.

_ Chrome and Firefox limit users to 6 concurrent requests per host_
6 concurrent requests should really be enough if our queries are in the ms. And it is really not relevant with websocket and http2 anyways.

_Another option would be to split the undecimated query stuff out into a separate route_
Yeah, I think they are different enough that it would make sense... Otherwise the API needs a lot of documentation, explaining for decimated queries you can specify fields a,b,c and for undecimated queries fields a,d,e...

tag-historian/nitaghistorian.yml
tschmittni: Maybe we should provide the type information outside of 'values' so that we just provide the info once and don't repeat it 1000 times.

pvallone: I agree that would save some space. It seems like we could provide a dictionary where the keys are tag paths and the values are types. That would be another breaking change, so I'll take a look at the impact that would have on tag viewer and push proposed changes for what that would look like.

tag-historian/nitaghistorian.yml
pvallone: > The client can just send the requests in parallel. And we can basically start displaying tiles when the first request gets returned and not when the slowest query finally finished.
This also makes it easier to apply throttling.

There are a lot of clients which would be impacted by such a change. This would break dashboard builder, and represent additional breakages to the G API and tag viewer. I would prefer the complexity of parallellizing queries to be in the service, rather than push that complexity to each of the clients of the API. If you still feel strongly about breaking the API in this way, I think we should have a broader discussion at the spec review meeting with all of the stakeholders who would be impacted.

> Yeah, I think they are different enough that it would make sense...

Carson tells me that dashboard builder doesn't do undecimated queries. I don't think tag viewer does either. The G API does, but I think we could handle that by just querying a different route in that case. I'll push changes to detail what the API might look like if the undecimated query stuff was split out into its own route.

You and the other SLC folks should also feel free to push to this branch if there's something you'd like to change, also.

pvallone: Another reason to continue to support specifying multiple tags is that this route is also used to export tag history to CSV files. If a user wants to export the history data of three tags to CSV, @prestwick's feeling is that they will want that data in one CSV file, not three. We don't want users, or our web UI clients, to have to make separate requests for each tag and then stitch the CSV files together.

pvallone: @spanglerco brought up the point that /v1/tags/query-history seems like a suitable path for *undecimated* queries, so it's actually the decimated queries that would be split out into a new route... I think I'm going to go with /v1/tags/query-decimated-history for now.

tschmittni: _There are a lot of clients which would be impacted by such a change._
everything what we discuss here is already breaking:
- removing skip
- adding continuation token support
- removing metadata
- removing complete routes
- etc...

I would say that if we already make these changes, then everything should be on the table to make this API more performant, concise and easier to understand.

If we don't want to break anybody than we have to keep the old API on SLS... and we just call this API /v2/. And maybe systemlink cloud will never implement /v1/.

It really depends on what we want to achieve. But I am just afraid that patching this API somewhat up while "partially" breaking it, might get us the disadvantages of both sides. We still end up patching all our clients and we might not be super happy about the results either...

tschmittni: _Another reason to continue to support specifying multiple tags is that this route is also used to export tag history to CSV files._

Doesn't need to be. If we want to support larger CSV downloads, we could provide a separate download API which is even async. And this can be used if clients want to perform data retention. what do you think?

spanglerco: > I would say that if we already make these changes, then everything should be on the table to make this API more performant, concise and easier to understand.

But there's a difference between an API break and changing the supported feature set. In this instance, Phil is referring to the impact of a feature loss on clients that are using that feature in a visible way.

> If we want to support larger CSV downloads, we could provide a separate download API which is even async.

No need to make it async when a CSV file can be streamed as it's generated. But the same use case applies to LabVIEW applications as well. What would be the difference between a download API and query-data? One would support multiple paths but not continuation token and the other would be single path with continuation token? I don't really see the benefit when the continuation token can be used to iterate over each tag path in turn.

pvallone:
@tschmittni I'll modify the swagger doc with what I think we proposed at the spec review:

- query-history (undecimated) takes one path and no longer supports CSV exports
- query-decimated-history continues to support multiple paths and no longer supports CSV exports
- A new CSV download route is created which does not support continuation tokens. This route supports querying for multiple tags, and it streams a single CSV file to the client.

pvallone:
I modified the swagger doc with changes from the spec review.

@tschmittni @spanglerco: I haven't modified the doc to break out undecimated and decimated queries in the web socket API. Do we want to split them out? That would sort of explode the web socket API, which is already pretty dense. We would need a lot of new types and actions:

- HistoricalTagDecimatedQuery
- HistoricalTagUndecimatedQuery
- Web Socket Decimated Query Request
- Web Socket Undecimated Query Request
- Web Socket Async Decimated Query Request
- Web Socket Async Undecimated Query Request

If we kept with the web socket API's current pattern of echoing back the query objects, we would also need: 
- Web Socket Decimated Query Response
- Web Socket Undecimated Query Response
- Web Socket Async Undecimated Query Response
- Web Socket Async Decimated Query Response
- HistoricalTagAsyncDecimatedQueryJob
- HistoricalTagAsyncUndecimatedQueryJob

tschmittni:
_Do we want to split them out?_

I would vote for keeping them in-sync... otherwise it might get really confusing

tschmittni:
_ I'll modify the swagger doc with what I think we proposed at the spec review:_

thank you!

pvallone:
> I would vote for keeping them in-sync... otherwise it might get really confusing

I split the async web socket query routes out. Paul's feeling was that nobody is using the synchronous web socket query API, so I removed those routes. I wasn't sure if we wanted the async undecimated query route to support querying for multiple tags or not. Since it's async, it seems like it could, so it currently does in my branch.

tag-historian/nitaghistorian.yml
```diff
@@ -731,187 +680,129 @@ paths:
...
-            properties:
-              tagHistorianValues:
-                $ref: '#/definitions/TagHistorianValues'
+                example: [system2.tag1]
         401:
           $ref: '#/responses/Unauthorized'
         default:
           $ref: '#/responses/Error'
-    put:
+  /v2/tags/download-csv:
```
tschmittni: how about calling the route: /v2/tags/export-history
and a parameter "format" with "csv"

So this is a little more extendable.

pvallone: "responseFormat" is what we had before, and that felt more clear, so I went with that.

pvallone:
@tschmittni Mark and Fred think that we should have separate swagger docs for the cloud and server historians, at least until both services have the same API. We'd like to check this swagger doc in (I renamed the file to nitaghistorian-cloud.yml) and pull it into api.systemlinkcloud.com. Could you approve this PR if you're satisfied with the API?

tschmittni (approved):

PR #57, opened 2019-07-19
Set timeout values in Swagger. Use the same values as in old WebService.

chotea:
Signed-off-by: Cristian Hotea <cristian.hotea@ni.com>

Set timeout values to routes that take more than the default 30000 ms to execute.


rares-pop (approved):
It's a bit strange that it only accepts static configuration, but oh well..

PR #59, opened 2019-07-29
Fix SystemManagement swagger

rares-pop:
- fix examples
- fix grains/feeds/packages, as dicts
- fix sysapi/groups and lists

Signed-off-by: Rares POP <rares.pop@ni.com>

### DIFF
--- a/systems-management/nisysmgmt.yml
+++ b/systems-management/nisysmgmt.yml
@@ -587,7 +587,25 @@ definitions:
         example: 19.0
     required: [activated]
 
-  ManagedSystemDataProperty:
+  ManagedSystemDataPropertyList:
+    description: Property of a system composed by the last updated timestamp and a list with the actual data.
+    type: object
+    properties:
+      lastUpdatedTimestamp:
+        description: Last updated timestamp.
+        type: string
+        format: iso-date-time
+        example: '2019-02-11T13:32:06.030958Z'
+      data:
+        description: The actual data.
+        type: array
+        items:
+          type: object
+          example: 
+            - Lab1
+    required: [lastUpdatedTimestamp, data]
+
+  ManagedSystemDataPropertyMap:
     description: Property of a system composed by the last updated timestamp and a dictionary with the actual data.
     type: object
     properties:
@@ -599,6 +617,8 @@ definitions:
       data:
         description: The actual data.
         type: object
+        additionalProperties:
+          type: object
         example:
           key: value
     required: [lastUpdatedTimestamp, data]
@@ -656,7 +676,7 @@ definitions:
         required: [lastUpdatedTimestamp, data]
       grains:
         description: General information about the system.
-        $ref: '#/definitions/ManagedSystemDataProperty'
+        $ref: '#/definitions/ManagedSystemDataPropertyMap'
         example:
           lastUpdatedTimestamp: '2019-02-19T11:42:25.078Z'
           data:
@@ -672,7 +692,7 @@ definitions:
             osrelease: 7        
       packages:
         description: Software packages installed on the system.
-        $ref: '#/definitions/ManagedSystemDataProperty'
+        $ref: '#/definitions/ManagedSystemDataPropertyMap'
         example:
           lastUpdatedTimestamp: '2019-02-19T11:42:25.078Z'
           data:
@@ -696,7 +716,7 @@ definitions:
               version: 19.5.0.28-0+d28
       feeds:
         description: Feeds configured on the system.
-        $ref: '#/definitions/ManagedSystemDataProperty'
+        $ref: '#/definitions/ManagedSystemDataPropertyMap'
         example:
           lastUpdatedTimestamp: '2019-02-19T11:42:25.078Z'
           data:
@@ -712,7 +732,7 @@ definitions:
               uri: http://download.ni.com/support/nipkg/products/ni-package-manager/eulas
       sysapi:
         description: System API information.
-        $ref: '#/definitions/ManagedSystemDataProperty'
+        $ref: '#/definitions/ManagedSystemDataPropertyList'
         example:
           lastUpdatedTimestamp: '2019-02-19T11:42:25.078Z'
           data:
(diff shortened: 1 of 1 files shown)


systems-management/nisysmgmt.yml
```diff
@@ -587,7 +587,25 @@ definitions:
...
+        type: string
+        format: iso-date-time
+        example: '2019-02-11T13:32:06.030958Z'
+      data:
+        description: The actual data.
+        type: array
+        items:
+          type: object
+          example: 
+            - Lab1
```
spanglerco: Did you intend to make the example an array inside the array? As in, this example is for an item of the array, not the data array itself.

systems-management/nisysmgmt.yml
```diff
@@ -587,7 +587,25 @@ definitions:
...
+      lastUpdatedTimestamp:
+        description: Last updated timestamp.
+        type: string
+        format: iso-date-time
+        example: '2019-02-11T13:32:06.030958Z'
+      data:
+        description: The actual data.
+        type: array
+        items:
+          type: object
```
spanglerco, line 615: So this can be an array of anything? There isn't any way to be more descriptive in the array's description for what clients can expect here (i.e. an array of scalars).

systems-management/nisysmgmt.yml
```diff
@@ -599,6 +617,8 @@ definitions:
       data:
         description: The actual data.
         type: object
+        additionalProperties:
+          type: object
```
spanglerco, line 631: Same comment as above: is there any restriction on the type of data that can be in this map that we could document?

systems-management/nisysmgmt.yml
```diff
@@ -734,11 +754,21 @@ definitions:
...
+            format: iso-date-time
+            example: '2019-02-19T11:42:25.078Z'
           data:
-            - 121b471c-d7d4-4996-a38f-27f7e58be070
+            description: A list of strings defining metadata information about a system.
+            type: array
+            items:
+              type: string
+              example: 
+                - Lab1
```
spanglerco: Same comment on whether you intended this example to be an array inside an array.

systems-management/nisysmgmt.yml
```diff
@@ -1119,10 +1149,9 @@ definitions:
         description: The name of the package.
         example: "ni-securityupdate-kb67l8lcqw-killbits"
       plugin:
-        type: integer
-        format: int32
+        type: string
         description: Agent that will be used to install the package.
-        example: 2
+        example: "wininst"
```
spanglerco, line 1162: Should this be an enum, or do we expect clients to be able to send custom values?

systems-management/nisysmgmt.yml
```diff
@@ -1131,10 +1160,9 @@ definitions:
           type: string
           example: "ni-teststand-2017-tools (>= 17.0.0)"
       priority:
-        type: integer
-        format: int32
+        type: string
         description: The priority of the package.
-        example: 4        
+        example: "standard"
```
spanglerco, line 1173: Ditto on enum vs. allowing custom.

spanglerco (changes requested):

systems-management/nisysmgmt.yml
rares-pop: Oops.. it was a copy-paste wrong indent problem for the example. However this definition is for SysAPI data, which is a list of property-bags (and we don't really care to define the property-bags, hence we let them 'objects'). 
I removed the example altogether since we don't really have a definition for the property-bags.
I see most of them have a resource_uri and an array of properties which have type/tag/value. However, our service doesn't consume any of those, so we take whatever sysAPI is publishing and return it as 'object', that way we won't have to change when sysAPI changes.

systems-management/nisysmgmt.yml
rares-pop: This is for sysapi data - I replied above why we use 'object' here.

systems-management/nisysmgmt.yml
rares-pop: This is used for grains, packages and feeds, and they are really heterogeneous collections.

Grains use strings, ints, bools, inner dictionaries or lists as elements. Also, grains are evolving as salt evolves, and their policy is - we use a dictionary and everyone is taking whatever it cares about from it, and that way there's no versioning.
Packages have a custom format that's defined by nipkg, which can have an open-ended collection of metadata, and feeds are similar too (they have different format of linux feeds, vs. windows (nipkg) feeds). 

So, there are really no constraints on those.

systems-management/nisysmgmt.yml
rares-pop: Oops.. copy-paste-wrong-identation case. I fixed it by inlining the example under the 'items', as opposed to un-indent it as it was.

systems-management/nisysmgmt.yml
rares-pop: Both this and the next are some changes Eliza did, and she said those are open-ended strings. Some of them are integers, some of them strings. She fixed this when she saw some exceptions for certain packages that contains priority = 'wininst' or 'standard'. It's best to let it a string, since we are not really interpreting those, but just read them from the system and feed them to nipkg or opkg.

systems-management/nisysmgmt.yml
rares-pop: I replied above.

systems-management/nisysmgmt.yml
spanglerco: That sounds like great information to include in the description instead of "The actual data." :)

systems-management/nisysmgmt.yml
spanglerco: Makes sense. Is there a way some of that information could be summarized in the description here?

rares-pop: I did add better descriptions.

systems-management/nisysmgmt.yml
rares-pop: I did change the description.

spanglerco (approved):

PR #69, opened 2020-02-21
Swagger document for WebApp Service

tschmittni:
Added Swagger document for /niapp routes of the WebApp Service.

This is an update to the old PR: https://github.com/ni/systemlink-OpenAPI-documents/pull/47

### DIFF
--- a/webapp/niapp.yaml
+++ b/webapp/niapp.yaml
@@ -36,8 +36,8 @@ paths:
     put:
       tags:
         - metadata
-      summary: Update existing webapp metadata
-      description: Update webapp metadata by Id
+      summary: Update webapp metadata by Id
+      description: Update existing webapp metadata
       operationId: update-webapp
       parameters:
         - $ref: '#/parameters/WebAppId'
@@ -54,8 +54,8 @@ paths:
     delete:
       tags:
         - metadata
-      summary: ''
-      description: Delete webapp by id
+      summary: Delete webapp by Id
+      description: Delete existing webapp metadata and content
       operationId: delete-webapp
       parameters:
         - $ref: '#/parameters/WebAppId'
@@ -70,8 +70,8 @@ paths:
     get:
       tags:
         - metadata
-      summary: Lists the webapps
-      description: Lists the webapps
+      summary: List the webapps
+      description: List the webapps
       operationId: list-webapps
       parameters:
         - in: query
@@ -88,16 +88,16 @@ paths:
           - WebVI
         - in: query
           name: userId
-          description: Filters the webapps by the user id of the owner
+          description: Filters the webapps by the user Id of the owner
           type: string
         - in: query
           name: workspace
-          description: Filters the webapps by workspace id
+          description: Filters the webapps by workspace Id
           type: string
         - in: query
           name: continuationToken
           description: The continuation token can be used to paginate through the webapp list results. Provide a token to continue a previous listing.
-          type: integer
+          type: string
         - in: query
           name: take
           description: How many webapps to return in the result
@@ -145,8 +145,8 @@ paths:
     get:
       tags:
         - metadata
-      summary: List the emails the user has shared webapps with
-      description: List the emails the user has shared webapps with
+      summary: List the emails with which the user has shared webapps
+      description: List the emails with which the user has shared webapps
       operationId: list-shared-emails
       responses:
         200:
@@ -173,7 +173,7 @@ paths:
       tags:
         - content
       summary: Upload content for an existing webapp
-      description: Upload content an existing webapp
+      description: Upload content for an existing webapp
       operationId: update-content
       parameters:
         - $ref: '#/parameters/WebAppId'
@@ -197,7 +197,7 @@ paths:
       tags:
         - content
       summary: Get webapp content by path
-      description: Get webapp content at path. For webapps with multiple files (e.g. WebVIs)
+      description: Get the webapp content at a path. This applies to webapps with multiple files, such as WebVIs.
       operationId: get-content
       parameters:
         - $ref: '#/parameters/WebAppId'
@@ -223,7 +223,7 @@ paths:
           type: string
           required: true
           name: sourceId
-          description: The id of the source webapp to get copy content from
+          description: The Id of the source webapp to get copy content from
       responses:
         200:
           description: Success
@@ -367,7 +367,7 @@ definitions:
     properties:
       id :
         type: string
-        description: The webapp id
+        description: The webapp Id
         example: "asdsad-17a6-45323-b64b-65325287372d"
       type:
         type: string
@@ -400,10 +400,12 @@ definitions:
         description: List of policy Ids associated with the webapp, which give it access to the user's resources"
       created:
         type: string
+        format: date-time
         description: The created timestamp (iso8601 format)
         example: "2019-12-02T15:31:45.379Z"
       updated:
         type: string
+        format: date-time
         description: The last updated timestamp (iso8601 format)
         example: "2019-12-02T15:31:45.379Z"
       properties:
--- a/webapp/niapp.yaml
+++ b/webapp/niapp.yaml
@@ -1,0 +1,416 @@
+swagger: '2.0'
+info:
+  version: '1.0'
+  title: SystemLink WebApp Service
+basePath: /niapp/v1
+consumes:
+  - application/json
+produces:
+  - application/json
+securityDefinitions:
+  ApiKeyAuth:
+    type: apiKey
+    name: x-ni-api-key
+    in: header
+security:
+  - ApiKeyAuth: []
+schemes:
+  - https
+paths:
+  '/webapps/{id}':
+    get:
+      tags:
+        - metadata
+      summary: Get webapp metadata by Id
+      description: Get webapp metadata by Id
+      operationId: get-webapp
+      parameters:
+        - $ref: '#/parameters/WebAppId'
+      responses:
+        200:
+          $ref: '#/responses/GetWebAppResponse'
+        401:
+          $ref: '#/responses/Unauthorized'
+        404:
+          $ref: '#/responses/NotFound'
+    put:
+      tags:
+        - metadata
+      summary: Update existing webapp metadata
+      description: Update webapp metadata by Id
+      operationId: update-webapp
+      parameters:
+        - $ref: '#/parameters/WebAppId'
+        - $ref: '#/parameters/UpdateWebAppRequest'
+      responses:
+        200:
+          $ref: '#/responses/UpdateWebAppResponse'
+        400:
+          $ref: '#/responses/ValidationError'
+        401:
+          $ref: '#/responses/Unauthorized'
+        404:
+          $ref: '#/responses/NotFound'
+    delete:
+      tags:
+        - metadata
+      summary: ''
+      description: Delete webapp by id
+      operationId: delete-webapp
+      parameters:
+        - $ref: '#/parameters/WebAppId'
+      responses:
+        200:
+          description: Success
+        401:
+          $ref: '#/responses/Unauthorized'
+        404:
+          $ref: '#/responses/NotFound'
+  '/webapps':
+    get:
+      tags:
+        - metadata
+      summary: Lists the webapps
+      description: Lists the webapps
+      operationId: list-webapps
+      parameters:
+        - in: query
+          name: name
+          description: Filters the webapps by name
+          type: string
+        - in: query
+          name: type
+          description: Filters the webapps by type
+          type: string
+          enum:
+          - Dashboard
+          - DashboardTemplate
+          - WebVI
+        - in: query
+          name: userId
+          description: Filters the webapps by the user id of the owner
+          type: string
+        - in: query
+          name: workspace
+          description: Filters the webapps by workspace id
+          type: string
+        - in: query
+          name: continuationToken
+          description: The continuation token can be used to paginate through the webapp list results. Provide a token to continue a previous listing.
+          type: integer
+        - in: query
+          name: take
+          description: How many webapps to return in the result
+          type: integer
+          default: 100
+          maximum: 1000
+      responses:
+        200:
+          $ref: '#/responses/ListWebAppsResponse'
+        401:
+          $ref: '#/responses/Unauthorized'
+    post:
+      tags:
+        - metadata
+      summary: Creates a new webapp
+      description: Creates a new webapp
+      operationId: create-webapp
+      parameters:
+        - $ref: '#/parameters/CreateWebAppRequest'
+      responses:
+        200:
+          $ref: '#/responses/CreateWebAppResponse'
+        401:
+          $ref: '#/responses/Unauthorized'
+  '/webapps/{id}/sharing':
+    put:
+      tags:
+        - metadata
+      summary: Update sharing settings of the webapp
+      description: Update sharing settings of the webapp
+      operationId: update-webapp-sharing
+      parameters:
+        - $ref: '#/parameters/WebAppId'
+        - $ref: '#/parameters/UpdateWebAppSharingRequest'
+      responses:
+        200:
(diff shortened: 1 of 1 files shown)


webapp/niapp.yaml
```diff
@@ -0,0 +1,416 @@
...
+          description: The webapp's content. JSON for Dashboards, HTML for WebVIs
+        401:
+          $ref: '#/responses/Unauthorized'
+        404:
+          $ref: '#/responses/NotFound'
+    put:
+      tags:
+        - content
+      summary: Upload content for an existing webapp
+      description: Upload content an existing webapp
```
mollykreis: Upload content _for_ an existing webapp

webapp/niapp.yaml
```diff
@@ -0,0 +1,416 @@
...
+          description: Filters the webapps by the user id of the owner
+          type: string
+        - in: query
+          name: workspace
+          description: Filters the webapps by workspace id
+          type: string
+        - in: query
+          name: continuationToken
+          description: The continuation token can be used to paginate through the webapp list results. Provide a token to continue a previous listing.
+          type: integer
```
mollykreis: According to the documentation, you pass an integer for the continuation token when making a request, but the response gives you a string continuation token. Is that correct? If so, that seems odd.

webapp/niapp.yaml
```diff
@@ -0,0 +1,416 @@
...
+        401:
+          $ref: '#/responses/Unauthorized'
+        404:
+          $ref: '#/responses/NotFound'
+  '/api/webapps/{id}/content/{path}':
+    get:
+      tags:
+        - content
+      summary: Get webapp content by path
+      description: Get webapp content at path. For webapps with multiple files (e.g. WebVIs)
```
mollykreis: Since there are two phrases here, it would be nice if this were written as complete sentence(s).

Maybe something like "Get the webapp content at a path. This applies to webapps with multiple files, such as WebVIs."

webapp/niapp.yaml
```diff
@@ -0,0 +1,416 @@
...
+
+  '/webapps/{id}/duplicate/{sourceId}':
+    post:
+      tags:
+        - content
+      summary: Duplicates webapp content
+      description: Duplicates webapp content
+      operationId: duplicate-content
+      parameters:
+        - $ref: '#/parameters/WebAppId'
```
mollykreis, line 221: From my understanding, this should be the name of the new webapp rather than the ID.

mollykreis (changes requested):
It doesn't look like the route for uploading webapps is documented here.

webapp/niapp.yaml
```diff
@@ -0,0 +1,416 @@
+swagger: '2.0'
+info:
+  version: '1.0'
+  title: SystemLink WebApp Service
+basePath: /niapp/v1
```
alexweav, line 5: @spanglerco will putting V1 in the basePath cause any issues with NI webserver on SystemLink Server?

If not, I actually really like putting the version in the basePath, especially given the precedent of using a separate swagger doc for different API versions. It seems to remove duplication a lot. We should just double-check that it will work correctly with NI web server, especially given that this implies a separate web service for each API version (basePath corresponds to web service).

webapp/niapp.yaml
```diff
@@ -0,0 +1,416 @@
...
+        200:
+          $ref: '#/responses/GetWebAppResponse'
+        401:
+          $ref: '#/responses/Unauthorized'
+        404:
+          $ref: '#/responses/NotFound'
+    put:
+      tags:
+        - metadata
+      summary: Update existing webapp metadata
```
alexweav: Nitpick, feel free to ignore: Usually the description is more specific than the summary, so it's weird that the summary is wordier than the description. They should probably either be the same, or the longer one should be in the description slot.

webapp/niapp.yaml
```diff
@@ -0,0 +1,416 @@
...
+        400:
+          $ref: '#/responses/ValidationError'
+        401:
+          $ref: '#/responses/Unauthorized'
+        404:
+          $ref: '#/responses/NotFound'
+    delete:
+      tags:
+        - metadata
+      summary: ''
```
alexweav: Missing summary

webapp/niapp.yaml
```diff
@@ -0,0 +1,416 @@
...
+      description: Delete webapp by id
+      operationId: delete-webapp
+      parameters:
+        - $ref: '#/parameters/WebAppId'
+      responses:
+        200:
```
alexweav, line 63: Generally the success response to a DELETE should be a 204 if the response is empty, and a 200 if the response contains other info.

webapp/niapp.yaml
```diff
@@ -0,0 +1,416 @@
...
+      operationId: create-webapp
+      parameters:
+        - $ref: '#/parameters/CreateWebAppRequest'
+      responses:
+        200:
+          $ref: '#/responses/CreateWebAppResponse'
+        401:
+          $ref: '#/responses/Unauthorized'
+  '/webapps/{id}/sharing':
+    put:
```
alexweav, line 126: Suggestion, feel free to ignore: For API completeness, we might also want to consider adding a GET /webapps/{id}/sharing in addition to the update route. Especially since the other properties below (shared-emails, content, etc.) have a corresponding GET route as well.

webapp/niapp.yaml
```diff
@@ -0,0 +1,416 @@
...
+        - $ref: '#/parameters/WebAppContentPath'
+      responses:
+        200:
+          $ref: '#/definitions/WebAppContent'
+        401:
+          $ref: '#/responses/Unauthorized'
+        404:
+          $ref: '#/responses/NotFound'
+
+  '/webapps/{id}/duplicate/{sourceId}':
```
alexweav, line 213: If we're duplicating an existing webapp, it seems like something like `/webapps/{existingWebappId}/duplicate` is a bit more fluent for this RPC call, and then possibly have name/id/other parameters for the duplicated one passed in the request body. As it's worded now, it seems like `duplicate` is being called on an ID which doesn't yet exist, which may be confusing.

webapp/niapp.yaml
```diff
@@ -0,0 +1,416 @@
...
+        name:
+          type: string
+          description: The webapp's name
+        policyIds:
+          type: array
+          items:
+            type: string
+            example: "asdsad-17a6-45323-b64b-65325287372d"
+          description: List of policy Ids associated with the webapp, which give it access to the user's resources"
+        properties:
```
alexweav, line 265: Do we want to consider adding keywords as well, to match Tag and other datatypes?

webapp/niapp.yaml
```diff
@@ -0,0 +1,416 @@
...
+          type: string
+          example: john.doe@ni.com
+        description: List of emails of users to share the webapp with. Applies when "shared" option is "direct"
+      policyIds:
+        type: array
+        items:
+          type: string
+          example: "asdsad-17a6-45323-b64b-65325287372d"
+        description: List of policy Ids associated with the webapp, which give it access to the user's resources"
+      created:
```
alexweav, line 401: OpenAPI supports an additional format specifier for timestamps stored as strings, `format: date-time`. Without this present, some code generators may not translate this field to a language-native datetime.

webapp/niapp.yaml
```diff
@@ -0,0 +1,416 @@
...
+        type: string
+        description: The created timestamp (iso8601 format)
+        example: "2019-12-02T15:31:45.379Z"
+      updated:
```
alexweav, line 406: A few other services have standardized on "createdAt" and "updatedAt" for these create/update timestamps.

webapp/niapp.yaml
tschmittni: No, this route requires an existing webapp (metadata) as source and target

webapp/niapp.yaml
tschmittni: This is the documentation of the existing implementation.

webapp/niapp.yaml
tschmittni: This is the documentation of the existing implementation.

webapp/niapp.yaml
tschmittni: This is the documentation of the existing implementation.

webapp/niapp.yaml
tschmittni: This is the documentation of the existing implementation.

webapp/niapp.yaml
tschmittni: This is the documentation of the existing implementation.

webapp/niapp.yaml
mollykreis: It doesn't look like that's what is happening when I click "duplicate" from the dashboards grid in SystemLink Cloud. This is the same route, right?

If I duplicate a dashboard named "123" so that the duplicated dashboard has a name of "123-copy", I see that a POST is made to "https://www.systemlinkcloud.com/api/webapps/123-copy/duplicate/0569ab99-94c4-4ce7-af87-f99ad041763e". I don't see any other call that is creating metadata.

webapp/niapp.yaml
spanglerco: No, it won't cause any problems. The only thing that we don't support is mixing variable and static text in a single segment (e.g. /v{version}).

webapp/niapp.yaml
mollykreis: @tschmittni, just wanted to make sure you saw the question I posted here.

tschmittni: The /api/webapps/... is a thin wrapper over the actual webapp service API. Most of the time, it is just a pass-through but in this case the duplicate webapp route wrapper does a few extra things, e.g. duplicate the policy of the webapp, get the webapp metadata and create a new webapp before it actually calls the duplicate content route of the webapp service

I think the documentation of this route here is accurate.

mollykreis (approved):

webapp/niapp.yaml
mollykreis: Ah, ok. I see the distinction now.

mollykreis (changes requested):
I recently found out that this is not the API exposed by SystemLink Cloud and that only /api/webapps/ routes are public for Cloud. In order to actually be able to leverage shared code between Server and Cloud in our clients, we are planning on creating a wrapper exposing the /api/webapps/ routes. This means that this Swagger document will not be a valid customer API for either platform, and we should not document it here in this public location. Instead, I think we should move this document to live within the webappservices repo as internal documentation.

If any Swagger document should live here, it would be the documentation for the /api/webapps routes. Carson is planning on starting to work on such a Swagger doc, and we plan to use that document to codegen part of our C# wrapper for Server.

tschmittni:
_I recently found out that this is not the API exposed by SystemLink Cloud and that only /api/webapps/ routes are public for Cloud_

/api/webapps is not a public API. This is just an internal route which is forwarded to the https://api.systemlinkcloud.com/niapp/v1 routes. 


_In order to actually be able to leverage shared code between Server and Cloud in our clients, we are planning on creating a wrapper exposing the /api/webapps/ routes._

You don't need to create another wrapper. The /api/webapps are internal cloud routes and not meant to be stable or used by external clients. They are also not following any SystemLink naming standards. They are just a wrapper around the public APIs. 

We should not create another wrapper on SystemLink server to leverage shared code. Instead of creating two separate wrappers providing the same abstraction, we should just use the existing common /niapp/v1 routes in both cases, Cloud and Server.

========

This could be really just a simple as this: https://github.com/ni-kismet/helium-webserver/pull/1171
And we move the dashboard builder over to these routes and remove the /api/webapps routes.

mure:
@tschmittni Thanks for clarifying this for us. One of the benefits of the /api/webapps wrapper in cloud is a simplified duplicate route. The /niapp/v1 webapps service only provides a route to duplicate content, meaning that our UIs have to make multiple requests just to duplicate a document. Would it make sense for us to add another route to the webapps service similar to the /api/webapps one?

webapp/niapp.yaml
```diff
@@ -0,0 +1,418 @@
...
+        400:
+          $ref: '#/responses/ValidationError'
+        401:
+          $ref: '#/responses/Unauthorized'
+        404:
+          $ref: '#/responses/NotFound'
+    delete:
+      tags:
+        - metadata
+      summary: Delete webapp by id
```
mollykreis: id -> Id

webapp/niapp.yaml
```diff
@@ -0,0 +1,418 @@
...
+      description: Deletes existing webapp metadata
```
mollykreis: I assume this also deletes any metadata associated with the web app as well?

webapp/niapp.yaml
```diff
@@ -0,0 +1,418 @@
...
+        401:
+          $ref: '#/responses/Unauthorized'
+        404:
+          $ref: '#/responses/NotFound'
+  '/webapps/shared-emails':
+    get:
+      tags:
+        - metadata
+      summary: List the emails the user has shared webapps with
+      description: List the emails the user has shared webapps with
```
mollykreis: List the emails with which the user has shared webapps

webapp/niapp.yaml
```diff
@@ -0,0 +1,418 @@
...
+          description: Success
+        401:
+          $ref: '#/responses/Unauthorized'
+        404:
+          $ref: '#/responses/NotFound'
+  '/webapps':
+    get:
+      tags:
+        - metadata
+      summary: Lists the webapps
```
mollykreis: You should be consistent in the verb conjugation used in the summaries & descriptions. For some routes it is something like "List the webapps" and other routes use "List**s** the webapps".

Looking at the swagger docs hosted on test.systemlink.io, it doesn't appear that we are consistent across documents, but we generally are consistent within a single document.

tschmittni:
_The /niapp/v1 webapps service only provides a route to duplicate content, meaning that our UIs have to make multiple requests just to duplicate a document. Would it make sense for us to add another route to the webapps service similar to the /api/webapps one?_

Sure, if you need it just go ahead.

mure:
The request body for the create/update metadata routes are missing parts of the model, such as the document type.

webapp/niapp.yaml
```diff
@@ -0,0 +1,418 @@
...
+    description: The webapp identifier
+    type: string
+    required: true
+  WebAppContentPath:
+    in: path
+    name: path
+    description: Path to webapp content
+    type: string
+    required: true
+  CreateWebAppRequest:
```
mure, line 248: The request body for the create/update metadata routes are missing parts of the model, such as the document type.

mure: addendum: the "shared" and "policyIds" parameters are required to create a webapp, but this isn't documented here.

tschmittni: Yeah, these parameters are currently required. But we moved updating the sharing parameters to a new route. And the shared parameters are getting removed:

https://github.com/ni-kismet/helium-webappservices/pull/374

webapp/niapp.yaml
```diff
@@ -0,0 +1,418 @@
...
+      responses:
+        200:
+          $ref: '#/responses/ListWebAppsResponse'
+        401:
+          $ref: '#/responses/Unauthorized'
+    post:
+      tags:
+        - metadata
+      summary: Creates a new webapp
+      description: Creates a new webapp
```
mollykreis: Creates -> Create

webapp/niapp.yaml
```diff
@@ -0,0 +1,418 @@
...
+          $ref: '#/responses/Unauthorized'
+        404:
+          $ref: '#/responses/NotFound'
+
+  '/webapps/{id}/duplicate/{sourceId}':
+    post:
+      tags:
+        - content
+      summary: Duplicates webapp content
+      description: Duplicates webapp content
```
mollykreis: Duplicates -> Duplicate

mollykreis (changes requested):

mollykreis (approved):

PR #71, opened 2020-03-18
Add documentation for query webapps route

mollykreis:

This PR adds documentation for the webapps query route.

Documentation is good.

I copied and pasted the document into editor.swagger.io and ensured everything looked correct.

### DIFF
--- a/webapp/niapp.yaml
+++ b/webapp/niapp.yaml
@@ -125,7 +125,7 @@ paths:
           $ref: '#/responses/CreateWebAppResponse'
         401:
           $ref: '#/responses/Unauthorized'
-  '/webapps/query-webapps':
+  '/webapps/query':
     post:
       tags:
         - metadata
@@ -133,7 +133,7 @@ paths:
       description: >-
         Use the Dynamic Linq query language to specify filters for webapps.
         An empty request body queries all webapps.
-      operationId: query-webapps
+      operationId: query
       parameters:
         - in: body
           name: postBody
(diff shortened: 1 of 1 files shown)


webapp/niapp.yaml
```diff
@@ -122,6 +125,29 @@ paths:
           $ref: '#/responses/CreateWebAppResponse'
         401:
           $ref: '#/responses/Unauthorized'
+  '/webapps/query-webapps':
```
tschmittni: Shouldn't that be just /webapps/query

/webapps/query-webapps seems redundant...

webapp/niapp.yaml
```diff
@@ -416,3 +442,21 @@ definitions:
...
     title: WebApp Content
     description: The webapp binary content. Depending on the webapp's type it can be a dashboard, template or *.nipkg file exported from LabVIEW NXG
+  WebAppsAdvancedQuery:
+    type: object
+    title: Advanced Query Object for WebApps
+    properties:
+      filter:
+        description: The webapp query filter in Dynamic Linq
+        type: string
+        example: name.Contains("myWebApp") || type == "WebVI"
```
tschmittni, line 500: Maybe we should list all the supported fields and operators. Otherwise it is trial and error for the users

mollykreis: Done. I based the format of the documentation on the APM swagger doc.

webapp/niapp.yaml
mollykreis: I've updated the documentation here. Once this is submitted, I'll also update the code in the webapp service.

tschmittni (dismissed):

webapp/niapp.yaml
```diff
@@ -416,3 +442,68 @@ definitions:
...
+      take:
+        description: The maximum number of webapps to return
+        type: integer
+        default: 100
+        minimum: 0
+        maximum: 1000
+        example: 10
+      continuationToken:
```
BKnight760, line 508: The continuationToken section should contain an "example" element, otherwise the continuation token will be displayed in the swagger editor as "string".  I know we don't plan on hosting this on the SystemLink server visible to users, so this is not a blocking issue.  If we do plan to host this eventually, several similar example elements will need to be fixed.

BKnight760 (dismissed):

webapp/niapp.yaml
mollykreis: I wasn't sure what example to use, so I used "token" to be consistent with the test monitor swagger doc.

BKnight760 (approved):

mollykreis:
@tschmittni, Matt noticed that there was an extraneous "api/" on the duplicate route. I removed it. Can you confirm that it was a mistake initially?
____
I was able to confirm this is the case, so I'm going to merge the PR.

PR #72, opened 2020-03-27
Add documentation for duplicate webapp route and error responses

mollykreis:

- Add documentation for the new duplicate webapp route that will be added to the webapp service.
- Add documentation for the error response format. (This documentation was adapted directly from the Alarm service)
- Update Dynamic LINQ support documentation

Documentation is good.

Verified the document renders correctly in the swagger editor.

### DIFF
--- a/webapp/niapp.yaml
+++ b/webapp/niapp.yaml
@@ -33,6 +33,8 @@ paths:
           $ref: '#/responses/Unauthorized'
         404:
           $ref: '#/responses/NotFound'
+        default:
+          $ref: '#/responses/Error'
     put:
       tags:
         - metadata
@@ -51,6 +53,8 @@ paths:
           $ref: '#/responses/Unauthorized'
         404:
           $ref: '#/responses/NotFound'
+        default:
+          $ref: '#/responses/Error'
     delete:
       tags:
         - metadata
@@ -66,6 +70,8 @@ paths:
           $ref: '#/responses/Unauthorized'
         404:
           $ref: '#/responses/NotFound'
+        default:
+          $ref: '#/responses/Error'
   '/webapps':
     get:
       tags:
@@ -112,6 +118,8 @@ paths:
           $ref: '#/responses/ValidationError'
         401:
           $ref: '#/responses/Unauthorized'
+        default:
+          $ref: '#/responses/Error'
     post:
       tags:
         - metadata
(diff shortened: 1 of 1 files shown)


webapp/niapp.yaml
```diff
@@ -235,7 +255,32 @@ paths:
...
+      parameters:
+        - in: path
+          type: string
+          required: true
+          name: sourceId
+          description: The Id of the source webapp to copy content from
+        - $ref: '#/parameters/DuplicateWebAppRequest'
+      responses:
+        200:
+          description: Success
```
scotia673: A successful request should include the newly-created webapp's metadata in the response right? I would expect the documentation to outline a response in the success case similar to the way the successful response is described for the create route.

scotia673 (dismissed):
Thanks for updating the success field, Molly!

webapp/niapp.yaml
```diff
@@ -235,7 +255,32 @@ paths:
           $ref: '#/responses/Unauthorized'
         404:
           $ref: '#/responses/NotFound'
+        default:
+          $ref: '#/responses/Error'
 
+  '/webapps/duplicate/{sourceId}':
```
spanglerco: At first I was thinking this should be a `POST /duplicate-webapps` to match our HTTP conventions, but then I realized sourceId is a path parameter and that there's already a `POST /webapps/{id}/duplicate/{sourceId}`. Would it makes sense to merge those two routes into a `POST /webapps/{id}/duplicate` where source ID is in the path and destination ID is optional in the body?

That way, the URL is implying the "duplicate" action on the `/webapps/{id}` resource.

mollykreis: We could do that. I'm not sure what it would take to drop support for the existing route, though. We would probably need to keep it for at least a little while. Do you have an issue keeping both routes for now?

spanglerco: No, that sounds fine to me.

mollykreis: After an offline conversation with @spanglerco, we decided there wasn't a pressing reason to try to have the new duplicate route be dual-purpose to handle specifying an ID to duplicate only the content or specifying a name + workspace to duplicate content + metadata. This simplifies the documentation, testing, and implementation.

I will, however, change this new route to be `/webapps/{sourceId}/duplicate`.

spanglerco (approved):

PR #80, opened 2020-06-02
Add workspace to nitaghistorian

tschmittni:


tag-historian/nitaghistorian.yml
```diff
@@ -147,6 +147,9 @@ definitions:
         example:
           - system1.tag1
           - system2.tag1
+      workspace:
+        description: The workspace Id for querying decimated tag values. If this is not specified, the tag values from the default workspace will be returned.
+        example: 0c80cf49-54e9-4e92-b117-3bfa574caa84
```
spanglerco: These all need to include the `type: string` since the default is object.

```suggestion
      workspace:
        description: The workspace Id for querying decimated tag values. If this is not specified, the tag values from the default workspace will be returned.
        type: string
        example: 0c80cf49-54e9-4e92-b117-3bfa574caa84
```

tag-historian/nitaghistorian.yml
```diff
@@ -226,6 +232,9 @@ definitions:
         description: The path of the tag corresponding to the historical value.
         type: string
         example: system1.tag1
+      workspace:
+        description: The workspace Id of the tag
+        example: 0c80cf49-54e9-4e92-b117-3bfa574caa84
```
spanglerco: Workspace also needs to be added to the output of the HTTP query routes (currently as part of HttpHistoricalValue, but there's a bug to move that up to the same level as type), right? Not sure if we would add it directly to the responses or to the TypeAndValues definition below.

spanglerco (changes requested):

tag-historian/nitaghistorian.yml
tschmittni: good catch

tag-historian/nitaghistorian.yml
tschmittni: I have added the workspace to the response type

spanglerco (approved):

PR #85, opened 2020-06-29
Move administrative routes that require whitelisted API keys under a different tag, at the end of list

gergo-papp:

- Move routes that require whitelisted keys at the end of the list and mark them as whitelisted operations

- These routes could lead to confusion because it's not obvious if customers can use them or not
- Customer have other ways of creating keys and using them to access resources (e.g. Serevr Administrators can issue permanent API Keys using the `POST niauth/v1/keys` route)

#### Example
image

### DIFF
--- a/auth/niauth.yaml
+++ b/auth/niauth.yaml
@@ -32,25 +32,6 @@ paths:
         default:
           $ref: '#/responses/Error'
 
-  /session-keys:
-    post:
-      tags:
-        - keys
-      summary: Creates temporary API Keys
-      description: Create a temporary session key which is valid for 1 hour. The caller needs a whitelisted API key to issue session keys.
-      operationId: create-session-key
-      parameters:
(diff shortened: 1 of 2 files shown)


spanglerco:
@BKnight760  is out this week, but I think I agree with him that we should still document the routes, but maybe tag them as internal (so they go into their own section, preferably at the bottom of the UI) with documentation indicating they require a whitelisted API key.

gergo-papp:
> @BKnight760 is out this week, but I think I agree with him that we should still document the routes, but maybe tag them as internal (so they go into their own section, preferably at the bottom of the UI) with documentation indicating they require a whitelisted API key.

I could add a new section/tag called `Whitelisted operations` at the end.

andreirogobete (approved):

spanglerco (approved):

